Issue #1, opened 2022-07-25
Rebrand everything


asportnoy:
- [x] Replace mentions of "Powercord" with "Replugged"
- [ ] #11

asportnoy:
Closing this, backend is a separate issue.

PR #8, opened 2022-07-25
TSify

lleyton:
This is a part of an ongoing effort to convert Replugged to TypeScript. I'll be working on more of the core components later. For now, here are the TypeScriptified injectors.


lexisother (commented):
What is the reason for making `envCheck` a function unlike the above import-to-evaluate module `elevate`?

lleyton:
Kept it like that because it was previously done that way. I'll change that and test.

injectors/index.ts
```diff
@@ -0,0 +1,127 @@
...
+      "\n"
+    );
+    console.log(
+      "Replugged wasn't able to inject itself due to missing permissions.",
+      "\n"
+    );
+    console.log("Try again with elevated permissions.");
+  } else {
+    console.error("fucky wucky", e);
+  }
```
asportnoy, line 126: Is this still necessary?

package.json
```diff
@@ -7,28 +7,31 @@
...
     "test": "jest",
     "lint": "eslint --ext js --ext jsx src",
-    "plug": "node injectors/index.js inject --no-exit-codes",
-    "unplug": "node injectors/index.js uninject --no-exit-codes",
-    "inject": "node injectors/index.js inject --no-exit-codes",
-    "uninject": "node injectors/index.js uninject --no-exit-codes"
+    "plug": "ts-node injectors/index.ts inject --no-exit-codes",
+    "unplug": "ts-node injectors/index.ts uninject --no-exit-codes",
+    "inject": "ts-node injectors/index.ts inject --no-exit-codes",
+    "uninject": "ts-node injectors/index.ts uninject --no-exit-codes"
```
asportnoy, line 13: I think it's better to not use ts-node and instead compile it. This will make it run faster since it doesn't need to be built every time.

package.json
```diff
@@ -7,28 +7,31 @@
...
   },
   "repository": "https://github.com/replugged-org/replugged.git",
   "author": "aetheryx & Cynthia K. Rey",
   "bugs": {
     "url": "https://github.com/replugged-org/replugged/issues"
   },
   "dependencies": {
+    "@types/node": "^18.6.1",
```
asportnoy, line 21: Move to devDependencies

asportnoy (changes requested):
You need to setup ESLint for typescript. While you're at it, add `root: true`.

package.json
lleyton: true, although the injector is only going to be run once, we have incremental complication enabled (not sure if ts-node takes advantage of that,) and the built-in update system isn't currently setup to rebuild the project IIRC (which is going to be something to do down the line)

injectors/index.ts
lleyton: Hmmm, good catch, just kept the code as almost verbatim, lol. In theory it shouldn't be, because we run the escalation script before, so I'll remove this. Might want to test on Linux regardless though.

package.json
asportnoy: Well if more of the project is going to be in TypeScript in the future then it will be run every time, so might as well get set up good now.

package.json
lleyton: Sure, I'll just do @types/node (and not typescript or ts-node, which I think is what you were asking to do in the first place,) as we probably want to include TypeScript as a Repliugged compiler in the future (similar to how we handle SCSS or JSX)

asportnoy: Keep TypeScript, just switch out ts-node

package.json
lleyton: hmmm, I guess, I already was planing to do it in a seperate PR

package.json
lleyton: Currently the updater only installs prod dependencies which breaks the build during the update process. We could either:
1. Keep @types/node as a normal dep
2. Have the updater also install dev deps

I prefer #1, as we probably want to have some dependencies that the updater shouldn't install by default, such as eslint.

lleyton:
Expanding the scope to the entire project, might as well

asportnoy:
Moving this to draft state until we're ready for it

Luna-devv:
This technically can be closed as the [rewrite](https://github.com/replugged-org/replugged/tree/rewrite) is written in typescript already

PR #9, opened 2022-07-26
Change related to the ToS explanation

unclamped:
I think the added phrase might be a bit vague, but the point gets through. I am willing to hear suggestions about how this can be possibly changed


asportnoy:
This definitely needs to be reworded. I can add it to the readme myself.

PR #16, opened 2022-07-26
fix context menu injections

Oocrop:


asportnoy:
What exactly is the issue?

Oocrop:
injectContextMenu couldn't find context menus because they got wrapped in another analytics wrappers

asportnoy:
This needs a rebase

asportnoy (approved):

PR #17, opened 2022-07-26
Hopefully fix another router issue

lexisother:
Not entirely sure about the semantics of that old `findByTree` usage, it was supposed to `forceUpdate` something far down the tree but the prop it searched for has vanished and so I have no way of knowing what it was trying to find.

### DIFF
--- a/src/Powercord/coremods/router/index.js
+++ b/src/Powercord/coremods/router/index.js
@@ -55,9 +55,9 @@ async function injectSidebar () {
 
 async function forceRouterUpdate () {
   // Views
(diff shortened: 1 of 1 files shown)


asportnoy:
Could you clarify on what the issue is and how I can reproduce it?

lexisother:
I'm not sure if user-facing functionality is impacted but anything that registers a route throws an exception due to that `findInTree` call returning `undefined`.

asportnoy:
This needs a rebase

lexisother:
Rebased.

Issue #18, opened 2022-07-26
[Plugin downloader] Specify branch


asportnoy:
Repo links support specifying a branch name by appending `/tree/<branchname>` eg `https://github.com/replugged-org/replugged/tree/main`. Have the plugin downloader support this syntax to check out a specific branch.

asportnoy:
Gotta manually close this bc I can't spell closes apparently 😂

PR #19, opened 2022-07-26
Downloader branch

asportnoy:
Cloes #18
@Strencher please review this

### DIFF
--- a/src/Powercord/plugins/pc-moduleManager/util/cloneRepo.js
+++ b/src/Powercord/plugins/pc-moduleManager/util/cloneRepo.js
@@ -1,6 +1,7 @@
 const { join } = require('path');
 const { spawn } = require('child_process');
 const fs = require('fs');
+const { REPO_URL_REGEX } = require('./misc');
 
 module.exports = async function download (url, powercord, type) {
   // const dir = type === 'plugin' ? join(__dirname, '..', '..') : join(__dirname, '..', '..', 'themes');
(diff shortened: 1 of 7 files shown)


Strencher (approved):
Looks good to me. Tested successfully as well.

src/Powercord/plugins/pc-moduleManager/util/misc.js
```diff
@@ -1,3 +1,5 @@
+exports.REPO_URL_REGEX = /https?:\/\/(?:www\.)?github\.com\/([^/\s]+)\/([^/\s]+)(?:\/tree\/(\S+))?/;
```
Vendicated, line 1:
```suggestion
exports.REPO_URL_REGEX = /https?:\/\/(?:www\.)?github\.com\/(\S+)\/(\S+)(?:\/tree\/(\S+))?/;
```

src/Powercord/plugins/pc-moduleManager/util/cloneRepo.js
```diff
@@ -13,12 +14,21 @@ module.exports = async function download (url, powercord, type) {
       dir = join(__dirname, '..', '..', '..', 'themes');
       break;
   }
-  console.log(dir);
-  const repoName = url.match(/([\w-]+)\/?$/)[1];
+
+  const urlMatch = url.match(REPO_URL_REGEX);
+  if (!urlMatch) {
+    console.error('Could not parse URL');
```
Vendicated, line 20: should probably log the actual url here, otherwise this log is pretty meaningless

src/Powercord/plugins/pc-moduleManager/util/misc.js
asportnoy: That also includes slashes though

Vendicated: oh I didn't see that. You don't need to include slashes though, you can use `\S+?`. ? makes it non greedy

asportnoy: Screenshot on 2022-07-26 at 12 44 04
Making it non-greedy causes it to not go all the way to the end.

PR #25, opened 2022-07-26
repair command

12944qwerty:
shortens 
```
npm run unplug
git pull
npm run plug
```
into `npm run repair`


asportnoy (approved):
Much better lol

D3XX3R (approved):
looks good

D-Brox:
I would have called it `replug`

asportnoy:
I think that would confuse people. `replug` sounds like either plug but for after you've already unplugged or some unknown branding term. Repair is more self-explanitory.

PR #27, opened 2022-07-27
Software Development Kit Lab

12944qwerty:
Taken from my [pc pr](https://github.com/powercord-org/powercord/pull/659)

My ToDo for the SDK right now is as follows, and I will add more later on
- [ ] Fix any existing features Cynthia added and add more, please suggest anything you wish to include
- [ ] Integrate [PowercordModuleFinder](https://github.com/VenPlugs/PowercordModuleFinder)
- [ ] Integrate [F8Fix](https://github.com/doggybootsy/F8Fix)

### DIFF
--- a/src/Powercord/plugins/pc-sdk/components/ForceUI.jsx
+++ b/src/Powercord/plugins/pc-sdk/components/ForceUI.jsx
@@ -1,8 +1,25 @@
+/* eslint-disable no-debugger */
 const { React, getModule } = require('powercord/webpack');
-const { ButtonItem } = require('powercord/components/settings');
-const { getOwnerInstance } = require('powercord/util');
+const { ButtonItem, KeybindRecorder } = require('powercord/components/settings');
+const { getOwnerInstance, findInTree } = require('powercord/util');
 
 class ForceUI extends React.PureComponent {
+  constructor (props) {
+    super(props);
+
+    this.state = {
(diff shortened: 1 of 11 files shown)


MulverineX:
Suggestion: RenderError component. For wrapping around your components to prevent crashes and provide info. 
[Example](https://github.com/AAGaming00/rich-quotes/blob/master/components/child/ErrorRender.jsx)

src/Powercord/plugins/pc-sdk/components/ForceUI.jsx
```diff
@@ -49,7 +49,7 @@ class ForceUI extends React.PureComponent {
       const discordTextarea = document.querySelector('form > div > div > div');
       const instance = getOwnerInstance(discordTextarea);
 
-      const ChannelTextAreaForm = instance?._reactInternals?.return?.return;
+      const ChannelTextAreaForm = instance?._reactInternals?.return?.return?.return?.return?.return;
```
lexisother, line 52: Why not use `findInReactTree`? This is bound to break, just like the old `children.props[0].children[2].props.children.props` garbage that is still everywhere.

lexisother: This was resolved over Discord.
image

PR #41, opened 2022-08-01
Add Extra Buttons to Plugin Cards

Wolfkid200444:
This was a suggestion by DracoClaw#0065


i18n/en-US.json
```diff
@@ -47,6 +47,9 @@
   "REPLUGGED_PLUGINS_UNINSTALL": "Uninstall {pluginCount, plural, =1 {plugin} other {plugins}}",
   "REPLUGGED_PLUGINS_UNINSTALL_SURE": "Are you sure you want to uninstall {pluginCount, plural, =1 {this plugin} other {these plugins}}?",
   "REPLUGGED_PLUGINS_VERSION": "Version",
+  "REPLUGGED_PLUGINS_GITHUB": "Github Repository",
```
FifiTheBulldog, line 50: GitHub has a capital H. Or just remove “GitHub”--I think “Repository” should suffice.

FifiTheBulldog (changes requested):
Looks good, just a teeny tiny tweak to the repo string and then it should be good to merge.

While I don’t really like the idea of the repo being specified manually in the manifest instead of being fetched automatically through Git, it makes more sense if we’re going to migrate to a store format at some point.

i18n/en-US.json
Wolfkid200444: sure

src/Powercord/plugins/pc-moduleManager/components/parts/BaseProduct.jsx
```diff
@@ -43,19 +46,47 @@ class BaseProduct extends React.PureComponent {
...
+              look={Button.Looks.LINK}
+              size={Button.Sizes.SMALL}
+              color={Button.Colors.TRANSPARENT}
+            > {Messages.REPLUGGED_PLUGINS_GITHUB}
+            </Button>
+          }
+
+          {
+            <Button
+              onClick={() => openExternal(this.props.Path)}
```
FifiTheBulldog, line 82: For some reason this is giving me `Error: Invalid URL` when I click on it. Finder does not open .

FifiTheBulldog (changes requested):
Strings look good. All buttons work except for “Open Folder” which gives me an error and does nothing when clicked. See my comment.

src/Powercord/plugins/pc-moduleManager/components/parts/BaseProduct.jsx
Wolfkid200444: I replace openExternal for openPath this should work hopefully
on commit [556b579](https://github.com/replugged-org/replugged/pull/41/commits/556b579d4364d44538f694a7d6627a7c57a5af74)

FifiTheBulldog (approved):
Everything seems to work as expected, as long as plugin developers format their Discord invites as codes and not URLs. LGTM.

Wolfkid200444:
awesome

PR #43, opened 2022-08-02
Add booster badge

asportnoy:
Screenshot on 2022-08-01 at 19 32 32
Could use more i18n but it works


LandenStephenss:
Doesn't the rest of the 18n files need the key for the badge.

LandenStephenss (approved):

PR #48, opened 2022-08-03
feat: lazy webpack

Vendicated:
Adds two new methods to powercord/webpack: subscribe and waitFor

These allow you to asynchronously wait for lazily loaded webpack modules and immediately run a callback once they are loaded. In other words, this is useful for patching lazy context menus, lazy modals, etc

An example usage would be
```js
require("powercord/webpack").subscribe(
    m => m?.default?.displayName === "BanConfirm",
    m => {
        require("powercord/injector").inject("lazy-banconfirm-patch", m, "default", () => "trolley");
    }
)
```

### DIFF
--- a/src/fake_node_modules/powercord/webpack/lazy.js
+++ b/src/fake_node_modules/powercord/webpack/lazy.js
@@ -1,0 +1,100 @@
+const listeners = new Map();
+let originalPush;
+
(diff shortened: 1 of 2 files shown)


lleyton (approved):
Looks good :3

asportnoy (approved):

PR #55, opened 2022-08-05
Fix module repo button for SSH git URLs

asportnoy:
Plugins/themes using SSH Git URLs (`git@github.com:username/reponame.git`) would not work when pressing the "Repository" button. This PR fixes that.


asportnoy:
@Vendicated approved [via Discord](https://discord.com/channels/1000926524452647132/1000955965304221728/1005273175057244262)
Screenshot on 2022-08-05 at 19 52 17

PR #57, opened 2022-08-06
Add IDs to product cards

hyblocker:
This PR adds `id` attributes to all buttons on product cards (i.e. Theme and Plugin cards) so that themes can easily customize them (say for instance, adding an icon next to the Discord invite button) without having to resort to hacky workarounds which usually have a tendency to break.


asportnoy (changes requested):
This should use classes and not IDs since the cards are use multiple times

asportnoy (approved):

Issue #31, opened 2022-07-28
Add check to see if user is already in the replugged discord server


Deepfried-Chips:
**Describe the feature request**
Add a check to see if user is already in the replugged discord server to prevent the banner at the top, which gets relatively annoying when installing replugged on multiple devices

**What Plugin would it be for?**
pc-notices

asportnoy:
This issue is available! If you would like to take it, please add a comment so I can assign it to you.

Navnedia:
@asportnoy I'll take on this issue if it's still available.

PR #60, opened 2022-08-07
New User Welcome Banner Conditions

Navnedia:
Addresses enhancement feature request #31 :
When a new user opens replugged for the first time (or after it's been freshly plugged), the welcome banner will only be displayed if the user isn't already in the replugged discord server.
This fix waits for the discord connection before checking because otherwise, the getGuilds() module doesn't function properly.

Let me know if I should refactor anything beforehand.


asportnoy (changes requested):
Please run the linter

asportnoy (approved):

PR #64, opened 2022-08-08
Move Themes and Quick CSS to separate settings pages

wingio:
Separates the themes and quick CSS into their own settings for ease of navigation and discovery.

image
image


lleyton (changes requested):
fix the linter checks

wingio:
should be fixed

lleyton (approved):
bet

PR #42, opened 2022-08-01
plugin/theme install deep links

Deepfried-Chips:
well technically a small http server using express
this closes feature request #13
usage is very straight forward
`http://localhost/6473/install?address=[git repo url]`
https://discord.com/channels/1000926524452647132/1000955965304221728/1003755868040474634 for usage demonstration

PR #66, opened 2022-08-09
Make plugin installer links open in-app

asportnoy:
Follow-up to #42 
`https://replugged.dev/install?url=...` links will open the install prompt instead of opening your browser.


lleyton (approved):
as long as you tested this locally, the code looks for merging

PR #67, opened 2022-08-09
Make the plugin downloader recognize replugged.dev installer links

asportnoy:
Acts just like if it were a GitHub link
Also refactored (again lol)


lleyton (approved):
nya

PR #68, opened 2022-08-09
Fix enabling experiments

FifiTheBulldog:
Fix the Discord experiments toggle. (Thank you Beef, I used the marked-up snippet here.) The only catch is that disabling experiments only sets the setting; it doesn’t actually disable them until the user reloads.

### DIFF
--- a/src/Powercord/plugins/pc-settings/experiments.js
+++ b/src/Powercord/plugins/pc-settings/experiments.js
@@ -1,0 +1,52 @@
+// Credit to Beefers for this snippet
+
+const { getModule } = require('powercord/webpack');
+
+let hasEnabled = false;
+
+async function enableExperiments () {
+  if (hasEnabled) {
+    return;
+  }
+
+  // Get the module used to get the current user, but not the one used elsewhere because discord™️
+  const userMod = getModule([ 'getUsers' ], false).__proto__;
+
+  // Get the connection open handler
+  const { CONNECTION_OPEN } = (await getModule([ 'isDeveloper' ]))._dispatcher._orderedActionHandlers;
+
+  function getHandler (handlerName) {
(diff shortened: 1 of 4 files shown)


src/Powercord/plugins/pc-settings/experiments.js
```diff
@@ -0,0 +1,52 @@
...
+
+  // Unpatch getCurrentUser
+  userMod.getCurrentUser = oldGCUser;
+
+  hasEnabled = true;
+}
+
+async function disableExperiments () {
+  hasEnabled = false;
+}
```
lexisother, line 47: Consider showing users a modal with info about a required reload before experiments are fully disabled.

lexisother (changes requested):

lleyton (approved):
nya

MichaelDavidGK:
image
Discord crashes

thennothinghappened:
what did you do to cause this?

MichaelDavidGK:
> what did you do to cause this?

just updated to latest which required reload

thennothinghappened:
not happening for me. I'll have a look but it's most likely one of your plugins

thennothinghappened:
hmm, not really sure what's going on. that's wrapped in a try catch and is supposed to error silently so it shouldn't be killing it.

MichaelDavidGK:
> not happening for me. I'll have a look but it's most likely one of your plugins

I'm removing all plugins and will be adding one by one to see if that's the issue, will take me a while cause I'm busy with something else rn (15 minutes?)

MichaelDavidGK:
Apparently it's PluginRepo from BD crashing it even tho nothing changed except Replugged update, welp I will live without it tbf

thennothinghappened:
weird, but good that's all cleared up now!

asportnoy:
Reverted this since it caused some issues and didn't work properly for everyone

PR #77, opened 2022-08-10
option to disable update toast

thennothinghappened:
allows you to disable the update toast (updates are still checked for, but it won't show a message about it unless you visit the updater)


lleyton (approved):
because first-world problems, am I right

Issue #88, opened 2022-08-10
Hide "Repository" button in Plugin cards if there's no git repository


asportnoy:
If a plugin doesn't have a git repository, the repository button in the settings card currently opens Replugged git repository. Instead, the button should just be hidden.
Discovered in #75

hummusbird:
Ill take this one!

hummusbird:
@asportnoy this can be closed now :P

asportnoy:
For next time, you can use closong keywords so it's closed automatically. https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #95, opened 2022-08-11
hide "Repository" button when no git info found

hummusbird:
I'd like someone to test this on MacOS or Linux 

using the getGitInfo func instead of existsSync wouldn't show any button because react moment


hummusbird:
fixes: #88

hummusbird:
fixes #88

Vendicated:
this is a very bad idea, you are using a sync fs (blocking) api inside a render method. That will lead to freezes / lag spikes. You should use getGitInfo or at least async fs. If the issue you had with getGitInfo is that it's async, you just have to use componentDidMount to start the call and set the state once it finished

hummusbird:
alr thanks, i try that

Vendicated (approved):

src/Powercord/plugins/pc-moduleManager/components/parts/BaseProduct.jsx
```diff
@@ -2,12 +2,23 @@ const { React, getModule, constants: { Routes }, i18n: { Messages } } = require(
 const { Divider, Button } = require('powercord/components');
 
 const { shell: { openExternal, openPath } } = require('electron');
-
+const Path = require('path');
```
lleyton, line 5: massive nitpick, but can we make this lowercase

lleyton (changes requested):

src/Powercord/plugins/pc-moduleManager/components/parts/BaseProduct.jsx
hummusbird: i only copied every other const ;-;

Vendicated: The other consts are React components, this one is a module 🦜

src/Powercord/plugins/pc-moduleManager/components/parts/BaseProduct.jsx
```diff
@@ -2,12 +2,23 @@ const { React, getModule, constants: { Routes }, i18n: { Messages } } = require(
...
 class BaseProduct extends React.PureComponent {
+  constructor (props) {
+    super(props);
+    this.state = {
+      gitInfo: null
+    };
+  }
+
+  async componentDidMount () {
+    await this.getGitInfo(this.props.Path).then(i => this.setState({ gitInfo: i }));
```
lleyton, line 19: same thing with the casing here, although I don't blame you because it was like this way before :/

could you change it anyways? (and references to this prop)

lleyton (changes requested):

lleyton (approved):

PR #99, opened 2022-08-11
Change settings store behaviour to support nested objects

Bricklou:
I updated how the setting store handles nested objects in files.

Before, with the old code, this following line:
```js
powercord.pluginManager.get("my-plugin").settings.set('a.b', true)
```
was saved as:
```json
{
  "a.b": true
}
```
now ! it is saved as:
```json
{
  "a": {
    "b": true
  }
}
```


Bricklou:
It's now backward compatible, it will convert all configs file using the old format into the new one

lleyton:
I think we already discussed this in the core chat, and tbh I don't see the point of including this.

PR #111, opened 2022-08-16
Add keys to InstalledProducts, fixes Cards opening wrong repos

Vendicated:
This issue happens because render returns an array of these without keys so react uses the index as key. Consequently, if you enter a search term, the items will be filtered and the indexes will shift which confuses react. Explicitly specifying the key fixes this issue

### DIFF
--- a/src/Powercord/coremods/moduleManager/components/manage/Plugins.jsx
+++ b/src/Powercord/coremods/moduleManager/components/manage/Plugins.jsx
@@ -10,6 +10,7 @@ class Plugins extends Base {
   renderItem (item) {
     return (
       <InstalledProduct
+        key={item.entityID}
         product={item.manifest}
         isEnabled={powercord.pluginManager.isEnabled(item.entityID)}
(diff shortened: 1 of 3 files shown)


src/Powercord/plugins/pc-moduleManager/components/manage/Themes.jsx
```diff
@@ -30,10 +30,11 @@ class Themes extends Base {
   renderItem (item) {
     return (
       <InstalledProduct
+        key={item.entityID}
         product={item.manifest}
         isEnabled={powercord.styleManager.isEnabled(item.entityID)}
         onToggle={async v => {
-          await this._toggle(item.entityID, v);
+          this._toggle(item.entityID, v);
```
Vendicated, line 37: (_toggle is not async so i removed the await)

Vendicated:
"Rebased"

lleyton (approved):

asportnoy (approved):

PR #132, opened 2022-08-19
Fix connections coremod

Penguin-Spy:
This fixes the coremod portion of the Connections API, which is responsible for making all registered connections appear in discord.
In order to be more resilient to discord updates, I chose to inject into the logic that determines what connections exist. The original implementation injected into the React components that display the connections, which broke whenever Discord rewrote the User Profile modal (which was like every other month).


lleyton (changes requested):
please fix the lint errors, and if someone could test this locally, that would be good

asportnoy:
Looks like ESLint doesn't support the nullish assignment operator, lemme see if I can update that

asportnoy:
Still some lint errors you have to fix

Penguin-Spy:
fixed the lint errors again, thx for updating the eslint config
for testing locally: my plugin [powercord-ppl-moe](https://github.com/Penguin-Spy/powercord-ppl-moe) and AAGaming00's [better-connections](https://github.com/AAGaming00/better-connections) both use this API.
somewhere along the line discord updated the format for icons that it expects, so better-connections' icons don't show; they'll need to fix that themselves (writing something to try and fix the icons ourselves seemed unnecessarily complex).
I have already updated my plugin to work with this API, so the icons should show.
if you have any questions about if something is working as expected, feel free to ping me on Discord (or leave a comment here too ig)

12944qwerty (approved):
Works for me.

lleyton (approved):
nya

PR #138, opened 2022-08-21
Fix updater bug

asportnoy:
A settings call was still using `this.settings` instead of `settings` when the updater failed.


lleyton (approved):
purrfect

PR #139, opened 2022-08-21
Install embed fixes

asportnoy:
Fix crash for build overrides
Fix non-plugin GH links not opening


lexisother (approved):

nares2024:
> Fix crash for build overrides Fix non-plugin GH links not opening

nares2024 (commented):
@nares2024

PR #28, opened 2022-07-27
Implement plugin embeds

lexisother:
Does as it says on the tin. WIP!

PR #141, opened 2022-08-22
Fix: "TypeError: Cannot read property 'match' of undefined"

spinfal:
## What is it?
This error comes from the recent [plugin-embeds](https://github.com/replugged-org/replugged/pull/28) addition, and it would cause Discord to crash. I'm not sure what in chat causes the error but I've managed to find a simple fix.

## Crash caused by plugins?
No, I removed all of my plugins. Doing so still gave the same result.

## Images
Crash GIF:
Crash GIF

DevConsole error:
DevConsole error


asportnoy:
Already fixed in #139

PR #142, opened 2022-08-23
[Plugin embed] Use GH username instead of manifest author

asportnoy:
Makes it harder for people to send malicious links


lexisother (approved):
Yop.

PR #146, opened 2022-08-24
Create reload command

bop34:
Pretty simple, this command reloads a plugin. Previously you had to type two commands, but this is slightly more convenient.


src/Powercord/coremods/moduleManager/commands/reload.js
```diff
@@ -0,0 +1,60 @@
...
+    } else if (isPlugin) { // Core internal plugin
+      return resp(false, `"${id}" provides core functionality and cannot be reloaded.`);
+    }
+
+    const manager = isPlugin ? powercord.pluginManager : powercord.styleManager;
+    if (!manager.isEnabled(id)) {
+      return resp(false, `"${id}" is already disabled.`);
+    }
+
+    manager.disable(id);
```
12944qwerty, line 24: Using `manager.remount(id)` will automatically remount it for plugins. This isn't the same as enabling and disabling as it 'takes it off' of discord and then back on.
You'll have to make a similar one for themes.

12944qwerty (approved):
Looks good from what I see. Someone test before merging

PR #147, opened 2022-08-24
User Folder

DracoClaw:
This PR moves the plugin, theme, and settings folders into a user subfolder. helping to better organize the structure, and clearly indicate that those are the only things end users should be touching.


asportnoy:
Please update the old migration to move the files to the new folder. This way we don't have to move them twice.

.gitignore
```diff
@@ -7,12 +7,12 @@ package-lock.json
 yarn.lock
 pnpm-lock.yaml
 
+user/*
 
-plugins/*
 !plugins/.gitkeep
```
asportnoy, line 12:
```suggestion
user/*
!user/*/.gitkeep
```
Also make a settings gitkeep while you're at it

.gitignore
```diff
@@ -7,12 +7,12 @@ package-lock.json
...
 
 src/Powercord/coremods/react-devtools/ext
-themes/*
+
 !themes/.gitkeep
```
asportnoy, line 16:
```suggestion
```

src/fake_node_modules/powercord/constants.js
```diff
@@ -7,7 +7,7 @@ module.exports = Object.freeze({
   REPO_URL: 'replugged-org/replugged',
 
   // Runtime
-  SETTINGS_FOLDER: join(__dirname, '..', '..', '..', 'settings'),
+  SETTINGS_FOLDER: join(__dirname, '..', '..', '..', 'user', 'settings'),
```
asportnoy, line 10: Let's set up a const for plugins and themes and use that

asportnoy (changes requested):

lexisother:
The migration script in the state it's in now won't work for people who have already migrated to the new directories.

asportnoy:
There are 2 scripts

lexisother:
Why not update the old one?

asportnoy:
The old one is for src/Powercord/x to user/x
The new one is for x to user/x

lexisother:
Right.

lleyton (approved):
hmmmm, something about modifying past migrations, especially if they've already been deployed, just really doesn't seem right (and there's a new migration anyways, so is it necessary?)

regardless, everything else seems okay, but I want someone to locally test this before merging

lleyton (changes requested):
users seem to be having issues with this
https://discord.com/channels/1000926524452647132/1000955965304221728/1012098224930168892

bop34:
Replugged doesn't work at all for me with this branch. It loads when starting Discord, but then returns
`An error occurred while initializing "settings.js"! Error: ENOENT: no such file or directory, mkdir 'C:\Users\[user]\powercord\user\settings'
    at mkdirSync (fs.js:987:3)
    at e.mkdirSync (electron/js2c/asar_bundle.js:5:12191)
    at Object.<anonymous> (C:\Users\[user]\powercord\src\Powercord\apis\settingsStore\store.js:8:3)
    at Object.<anonymous> (C:\Users\[user]\powercord\src\Powercord\apis\settingsStore\store.js:118:3)
    at Module._compile (internal/modules/cjs/loader.js:1078:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1108:10)
    at Module.load (internal/modules/cjs/loader.js:935:32)
    at Module._load (internal/modules/cjs/loader.js:776:14)
    at Function.f._load (electron/js2c/asar_bundle.js:5:12913)
    at Function.o._load (electron/js2c/renderer_init.js:33:379)`
and
`Uncaught (in promise) TypeError: Cannot read property 'buildCategoryObject' of undefined
    at Powercord.startup (C:\Users\[user]\powercord\src\Powercord\index.js:119:44)
    at async Powercord.init (C:\Users\[user]\powercord\src\Powercord\index.js:90:5)`

lexisother:
That's because the directory isn't being created recursively. `user` does not exist so it cannot create `settings`.

asportnoy:
> hmmmm, something about modifying past migrations, especially if they've already been deployed, just really doesn't seem right (and there's a new migration anyways, so is it necessary?)
> 
> regardless, everything else seems okay, but I want someone to locally test this before merging

I think in this case, it's alright. But if we wanted to change it, we could either:
- Rename the original one to make it a "new" migration
- Merge it into the new migration

.gitignore
asportnoy: @DracoClaw doesn't look like the gitkeep files were actually commited, can you double check?

DracoClaw: should be there now

lleyton:
@asportnoy I think the option of renaming the original to make it into a "new" migration would work

src/Powercord/migrations/20220824-move-user-items.js
```diff
@@ -0,0 +1,32 @@
+const fs = require('fs');
+const { join } = require('path');
+
+module.exports = () => {
+  const oldPlugs = join(__dirname, '..', '..', '..', 'plugins');
+  const newPlugs = join(__dirname, '..', '..', '..', 'user', 'plugins');
+  const oldThemes = join(__dirname, '..', '..', '..', 'themes');
+  const newThemes = join(__dirname, '..', '..', '..', 'user', 'themes');
+  const oldSettings = join(__dirname, '..', '..', '..', 'settings');
+  const newSettings = join(__dirname, '..', '..', '..', 'user', 'settings');
```
asportnoy, line 10: Use the constants for the new directories

asportnoy (changes requested):
Make the quick css changes as discussed in Discord

asportnoy:
Closing per poll in Discord

PR #149, opened 2022-08-25
Add Flatpak installation support

V3L0C1T13S:
Adds support for Flatpak installations of Discord and Discord Canary.

### DIFF
--- a/injectors/linux.js
+++ b/injectors/linux.js
@@ -7,12 +7,18 @@ const { BasicMessages, AnsiEscapes, PlatformNames } = require('./log');
 // This is to ensure the homedir we get is the actual user's homedir instead of root's homedir
 const homedir = execSync('grep $(logname) /etc/passwd | cut -d ":" -f6').toString().trim();
 
+const flatpakDir = '/var/lib/flatpak/app/com.discordapp';
+
+const homeFlatpakDir = `${homedir}/.local/share/flatpak/app/com.discordapp`;
+
 const KnownLinuxPaths = Object.freeze({
   stable: [
     '/usr/share/discord',
     '/usr/lib64/discord',
     '/opt/discord',
     '/opt/Discord',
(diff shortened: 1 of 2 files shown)


lleyton (changes requested):
Maybe we could run the flatpak override command in the installer?

Vendicated:
I don't think we should mess with people's settings without their consent, so it's better to tell the user to do it themselves. In addition to that, the replugged directory *might* be a symlink, in which case the user also has to grant permissions to the linked to folder, so in that case it would be confusing.

As for updating replugged and plugins, I included a shell script to do so manually in a similar PR I made to powercord, maybe something like that could be included here. For example, it could advise the user to run it on a chron job or modify their discord launcher to run it every time they launch discord
https://github.com/powercord-org/powercord/pull/485

lleyton:
Fair enough. On the point of updating Replugged and plugins, I would be open to including that shell script as an npm script, a temporary stopgap. I think a better long-term solution is to:
1. Ship replugged as a Flatpak and install the Git binary, see [Luna's comment](https://github.com/powercord-org/powercord/pull/485#issuecomment-951066235) on this.
2. Eventually remove the dependency on the Git binary, whether that's through WASM or isomorphic-git.

Vendicated:
I like those ideas! The shell script was more of a temporary solution since in powercords state at that time, there was no way to realise either of those ideas, but both of those would be nice solutions. For solution 1 there would just have to be someone willing to maintain it. Especially updating the canary one would probably be a chore

lleyton:
it's possible that we could have a bot to bump the Discord version, if that's the main burden with maintaining a Replugged flatpak

lleyton (changes requested):
Lovely! Could we add a message of how to update powercord/plugins (until we get something more proper?) Maybe just reference Ven's script?

lleyton (approved):
LGTM

PR #152, opened 2022-08-25
Fix an oversight with the updater regex

lexisother:


src/fake_node_modules/powercord/entities/Updatable.js
```diff
@@ -122,7 +122,7 @@ class Updatable extends Events {
       return await PowercordNative.exec('git remote get-url origin', {
         cwd: this.entityPath,
         timeout: TIMEOUT
-      }).then((r) => r.stdout.toString().match(/github\.com[:/]([\w-_]+\/[\w-_]+)/)[1] ||
+      }).then((r) => r.stdout.toString().match(/github\.com[:/]([\w-_]+\/[\w-_]+)/)?.[1] ||
           r.stdout.toString().trim().match(/(.*):(.*\/.*)/)[2]);
```
asportnoy, line 126: Should do the same here

lexisother: crap, I forgot there were two occurrences

asportnoy (approved):

PR #154, opened 2022-08-26
Memorize install directory on Linux

oatmealine:
Port of [powercord-org/powercord#602](https://github.com/powercord-org/powercord/pull/602)'s first half (as the second half seems to have already been implemented, separately):

> Currently, if a user is using a non-standard Discord install location, they will have to enter it each time they inject and uninject Powercord. This PR will fix this by creating an `.installdir` file and using that if possible.

Seperate `.installdir` files will be made for different Discord branches - for instance, Canary will have `.installdir-canary` and PTB will have `.installdir-ptb`

### DIFF
--- a/injectors/linux.js
+++ b/injectors/linux.js
@@ -1,11 +1,13 @@
 const { join } = require('path');
-const { existsSync } = require('fs');
+const { existsSync, readFileSync, writeFileSync } = require('fs');
 const { execSync } = require('child_process');
 const readline = require('readline');
 const { BasicMessages, AnsiEscapes, PlatformNames } = require('./log');
 
 // This is to ensure the homedir we get is the actual user's homedir instead of root's homedir
 const homedir = execSync('grep $(logname) /etc/passwd | cut -d ":" -f6').toString().trim();
+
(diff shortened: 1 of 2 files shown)


injectors/linux.js
```diff
@@ -1,12 +1,14 @@
...
-const { existsSync } = require('fs');
+const { existsSync, readFileSync, writeFileSync } = require('fs');
 const { execSync } = require('child_process');
 const readline = require('readline');
 const { BasicMessages, AnsiEscapes, PlatformNames } = require('./log');
 
 // This is to ensure the homedir we get is the actual user's homedir instead of root's homedir
 const homedir = execSync('grep $(logname) /etc/passwd | cut -d ":" -f6').toString().trim();
 
+const installDirFile = __dirname + '/../.installdir-';
```
asportnoy, line 10: path.join preferred for this

asportnoy (commented):
LGTM for the most part

injectors/linux.js
oatmealine: Should be good as of 86e558b

lleyton (approved):
LGTM, nya

PR #155, opened 2022-08-26
add plugin settings category

12944qwerty:
Adds the plugin category

Untested as I just coded this on my chromebook

### DIFF
--- a/src/Powercord/coremods/settings/index.js
+++ b/src/Powercord/coremods/settings/index.js
@@ -47,14 +47,26 @@ async function patchSettingsComponent () {
 
     const changelog = sections.find(c => c.section === 'changelog');
     if (changelog) {
-      const settingsSections = Object.keys(powercord.api.settings.tabs).map(s => _makeSection(s));
+      const coreSections = Object.keys(powercord.api.settings.tabs)
+        .filter(s => ['pc-commands', 'pc-i18n', 'pc-moduleManager', 'pc-notices', 'pc-settings', 'pc-updater'].indexOf(s) >= 0)
+        .map(s => _makeSection(s));
+      const pluginSections = Object.keys(powercord.api.settings.tabs)
+        .filter(s => ['pc-commands', 'pc-i18n', 'pc-moduleManager', 'pc-notices', 'pc-settings', 'pc-updater'].indexOf(s) < 0)
+        .map(s => _makeSection(s));
+      
       sections.splice(
         sections.indexOf(changelog), 0,
         {
           section: 'HEADER',
           label: 'Replugged'
         },
-        ...settingsSections,
+        ...coreSections,
+        { section: 'DIVIDER' },
+        {
+          section: 'HEADER',
+          label: 'Plugins'
+        },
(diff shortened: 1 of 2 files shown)


asportnoy (changes requested):
- General Settings, Plugins, Themes, and Quick CSS are all showing in the plugins settings section
- Make the section header an i18n key and name it "Plugin Settings" instead of just "Plugins"
- Sort by the plugin name instead of ID; some of them are visually out-of-order since the ID is different from the name.

src/Powercord/coremods/settings/index.js
```diff
@@ -48,11 +48,11 @@ async function patchSettingsComponent () {
...
     if (changelog) {
       const coreSections = Object.keys(powercord.api.settings.tabs)
-        .filter(s => ['pc-commands', 'pc-i18n', 'pc-moduleManager', 'pc-notices', 'pc-settings', 'pc-updater'].indexOf(s) >= 0)
+        .filter(s => ['pc-moduleManager-themes', 'pc-moduleManager-css', 'pc-moduleManager-plugins', 'pc-general', 'pc-updater'].indexOf(s) >= 0)
         .map(s => _makeSection(s));
       const pluginSections = Object.keys(powercord.api.settings.tabs)
-        .filter(s => ['pc-commands', 'pc-i18n', 'pc-moduleManager', 'pc-notices', 'pc-settings', 'pc-updater'].indexOf(s) < 0)
-        .sort()
+        .filter(s => ['pc-moduleManager-themes', 'pc-moduleManager-css', 'pc-moduleManager-plugins', 'pc-general', 'pc-updater'].indexOf(s) < 0)
+        .sort((a, b) => powercord.api.settings.tabs[a].label < powercord.api.settings.tabs[b].label)
```
lleyton, line 55: let's use [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) instead

lleyton (changes requested):

src/Powercord/coremods/settings/index.js
```diff
@@ -47,14 +47,27 @@ async function patchSettingsComponent () {
 
     const changelog = sections.find(c => c.section === 'changelog');
     if (changelog) {
-      const settingsSections = Object.keys(powercord.api.settings.tabs).map(s => _makeSection(s));
+      const coreSections = Object.keys(powercord.api.settings.tabs)
+        .filter(s => ['pc-moduleManager-themes', 'pc-moduleManager-css', 'pc-moduleManager-plugins', 'pc-general', 'pc-updater'].indexOf(s) >= 0)
```
asportnoy, line 51: Make this array a variable to avoid repetition 
Use `Array.includes` instead of `indexOf`

asportnoy (changes requested):

asportnoy:
Also, the linter needs to be run

asportnoy (approved):

Vendicated (approved):

PR #167, opened 2022-08-30
Fix crash if you have no plugins

asportnoy:
Tested locally


12944qwerty (approved):
lgtm as long as it works

PR #173, opened 2022-08-31
Update injector's platforms handling

### LINKED ISSUE #172: Update injector bad process handling
Only check one arg, not all of them
If a platform is specified but not valid, show an error and return
If no platform is specified, show a warning that it's defaulting to stable and continue

milena-kos:
Closes #172


lleyton (approved):
fuck it ship it

PR #177, opened 2022-09-01
Auto detect platform if none was specified

elementemerald:
This seems like a useful feature to have, avoiding the need to keep typing your platform every time you want to un/plug. Running `npm run un/plug` will try un/injecting into the following order of platforms: `stable -> ptb -> canary -> dev`. You would only need to type your platform if you have multiple platforms installed, so if you have only one, this should eliminate the process. Let me know if there's anything I should change.


injectors/index.js
```diff
@@ -48,13 +48,38 @@ if (!VALID_PLATFORMS.includes(platform)) {
...
   if (process.argv[2] === 'inject') {
-	try {
-		result = await main.inject(platformModule, platform);
-	} catch (e) {
-		// this runs if path generator crashes (app folder doesnt exist)
-		console.log(`${AnsiEscapes.RED}Platform you specified isn't installed on this device!${AnsiEscapes.RESET}\n\nList of valid platforms:\n${AnsiEscapes.GREEN}${VALID_PLATFORMS.map(x => `${x}`).join('\n')}${AnsiEscapes.RESET}`);
-		process.exit(process.argv.includes('--no-exit-codes') ? 0 : 1);
-	}
+    try {
+      if (!process.argv.find(x => VALID_PLATFORMS.includes(x.toLowerCase()))) {
```
asportnoy, line 52:
```suggestion
      if (!VALID_PLATFORMS.includes(process.argv[3]?.toLowerCase())) {
```
Should've been done in #172

injectors/index.js
```diff
@@ -48,13 +48,38 @@ if (!VALID_PLATFORMS.includes(platform)) {
...
+        for (const current of VALID_PLATFORMS) {
+          try
+          {
+            result = await main.inject(platformModule, current);
```
asportnoy, line 56: Instead of just trying to inject, add a function to check if the platform exists without installing, then add a log saying installing to {platform}.

injectors/index.js
```diff
@@ -73,13 +98,29 @@ List of valid platforms:\n${AnsiEscapes.GREEN}${VALID_PLATFORMS.map(x => `${x}`)
...
   } else if (process.argv[2] === 'uninject') {
     try {
-		result = await main.uninject(platformModule, platform);
-	} catch (e) {
-		// this runs if path generator crashes (app folder doesnt exist)
-		console.log(`${AnsiEscapes.RED}Platform you specified isn't installed on this device!${AnsiEscapes.RESET}\n\nList of valid platforms:\n${AnsiEscapes.GREEN}${VALID_PLATFORMS.map(x => `${x}`).join('\n')}${AnsiEscapes.RESET}`);
-		process.exit(process.argv.includes('--no-exit-codes') ? 0 : 1);
-	}
-	if (result) {
+      if (!process.argv.find(x => VALID_PLATFORMS.includes(x.toLowerCase()))) {
```
asportnoy, line 101: Let's make a shared function to determine the platform so we don't have to rewrite the same code.

asportnoy (changes requested):

asportnoy (changes requested):
There are still unaddressed issues from the last review

elementemerald:
I have addressed the above issues.

injectors/index.js
asportnoy: This is not handled yet

elementemerald: I made a shared function to handle this already on line 48 for both un/injecting.

asportnoy: https://github.com/replugged-org/replugged/blob/bd20ab37ab91508278aceb409694b1446f0301e7/injectors/index.js#L31

elementemerald: Should be handled now

lleyton (approved):
LGTM

injectors/index.js
asportnoy: Still not

elementemerald: Don't understand what still needs fixing here?

asportnoy: You're checking all the args. Only one should be checked.

elementemerald: Fixed

injectors/index.js
```diff
@@ -27,34 +27,56 @@ try {
...
+  if (platform) {
+    if (!exists) {
+      console.log(`${AnsiEscapes.RED}Platform you specified isn't valid, please specify a valid one.${AnsiEscapes.RESET}\n\nList of valid platforms:\n${AnsiEscapes.GREEN}${main.VALID_PLATFORMS.map(x => `${x}`).join('\n')}${AnsiEscapes.RESET}`);
+      process.exit(process.argv.includes('--no-exit-codes') ? 0 : 1);
+    } else if (platform === 'development') {
+      platform = 'dev';
+    }
   } else {
-    platform = platform.toLowerCase();
+    platform = 'stable';
```
asportnoy, line 43: Move code to auto detect platform to here
When checking, don't plug, just check if it's there.

asportnoy (changes requested):

asportnoy (changes requested):
There's some miscommunication going on here. BEFORE you do anything with injecting or trying to inject or anything like that, you should be able to determine which platform you're using.
https://github.com/replugged-org/replugged/blob/463580c9671890220abcbb455d9dc5b21c23ccf1/injectors/index.js#L30
The code to determine the platform should be right below this. If there's a platform specified and it's invalid, log that and exit. If there's no platform specified, find the first installed platform, log that, and continue. You shouldn't need to do any changes after that point.

injectors/index.js
```diff
@@ -27,34 +27,44 @@ try {
...
+    if (!exists) {
+      console.log(`${AnsiEscapes.RED}Platform you specified isn't valid, please specify a valid one.${AnsiEscapes.RESET}\n\nList of valid platforms:\n${AnsiEscapes.GREEN}${main.VALID_PLATFORMS.map(x => `${x}`).join('\n')}${AnsiEscapes.RESET}`);
+      process.exit(process.argv.includes('--no-exit-codes') ? 0 : 1);
+    } else if (platform === 'development') {
+      platform = 'dev';
+    }
   } else {
-    platform = platform.toLowerCase();
+    for (const current of main.VALID_PLATFORMS) {
+      if (current !== 'development') {
```
asportnoy, line 43: Why not development?

injectors/index.js
```diff
@@ -27,34 +27,44 @@ try {
...
+        try {
+          const installed = await main.checkInstalled(platformModule, current);
+          if (installed) {
+            console.log(`${AnsiEscapes.GREEN}${current} is installed, attempting to ${process.argv[2] === 'uninject' ? 'unplug' : 'plug'}.${AnsiEscapes.RESET}`);
```
asportnoy, line 47: Something like "No platform specified, defaulting to {platform}" would be better

injectors/index.js
```diff
@@ -27,34 +27,44 @@ try {
...
+            platform = current;
+            break;
+          } else {
+            console.log(`${AnsiEscapes.RED}${current} is not installed, skipping.${AnsiEscapes.RESET}`);
+          }
+        } catch (e) {
+          console.log(`${AnsiEscapes.RED}${current} is not installed, skipping.${AnsiEscapes.RESET}`);
```
asportnoy, line 54: Don't log that you're skipping it.

injectors/index.js
```diff
@@ -27,34 +27,44 @@ try {
...
+        }
+      }
+    }
   }
```
asportnoy, line 59: If platform doesn't exist here, say "Could not find any installations of Discord" and exit

injectors/index.js
```diff
@@ -27,34 +27,44 @@ try {
...
-		console.log(`${AnsiEscapes.RED}Platform you specified isn't installed on this device!${AnsiEscapes.RESET}\n\nList of valid platforms:\n${AnsiEscapes.GREEN}${VALID_PLATFORMS.map(x => `${x}`).join('\n')}${AnsiEscapes.RESET}`);
-		process.exit(process.argv.includes('--no-exit-codes') ? 0 : 1);
-	}
+    try {
+      result = await main.inject(platformModule, platform);
+    } catch (e) {
+      // this runs if path generator crashes (app folder doesnt exist)
+      console.log(`${AnsiEscapes.RED}Platform you specified isn't installed on this device!${AnsiEscapes.RESET}\n\nList of valid platforms:\n${AnsiEscapes.GREEN}${main.VALID_PLATFORMS.map(x => `${x}`).join('\n')}${AnsiEscapes.RESET}`);
+      process.exit(process.argv.includes('--no-exit-codes') ? 0 : 1);
+    }
```
asportnoy, line 67: Make sure to run the linter

injectors/index.js
```diff
@@ -68,24 +78,24 @@ if (!VALID_PLATFORMS.includes(platform)) {
...
-		process.exit(process.argv.includes('--no-exit-codes') ? 0 : 1);
-	}
-	if (result) {
+      result = await main.uninject(platformModule, platform);
+    } catch (e) {
+      // this runs if path generator crashes (app folder doesnt exist)
+      console.log(`${AnsiEscapes.RED}Platform you specified isn't installed on this device!${AnsiEscapes.RESET}\n\nList of valid platforms:\n${AnsiEscapes.GREEN}${main.VALID_PLATFORMS.map(x => `${x}`).join('\n')}${AnsiEscapes.RESET}`);
+      process.exit(process.argv.includes('--no-exit-codes') ? 0 : 1);
+    }
+    if (result) {
```
asportnoy, line 92: Make sure to run the linter

injectors/main.js
```diff
@@ -87,3 +89,12 @@ exports.uninject = async ({ getAppDir }, platform) => {
...
 };
+
+exports.checkPlatform = (platform) => {
+  return this.VALID_PLATFORMS.includes(platform);
+};
+
+exports.checkInstalled = async ({ getAppDir }, platform) => {
+  const appDir = await getAppDir(platform);
+  return existsSync(join(appDir, '..'));
+};
```
asportnoy, line 100: Let's move these back to index and just put them inline. Move the const above too. I don't see any need for them to be in this file.

asportnoy (changes requested):
Much better, just a few small things

injectors/index.js
```diff
@@ -29,32 +30,57 @@ try {
...
+        }
+      } catch (e) {
+        //
+      }
+    }
+
+    if (!platformFound) {
+      console.log(`${AnsiEscapes.RED}Could not find any installations of Discord.${AnsiEscapes.RESET}`);
+      process.exit(process.argv.includes('--no-exit-codes') ? 0 : 1);
+    }
```
asportnoy, line 69:
```suggestion
    for (const current of VALID_PLATFORMS) {
      try {
        const appDir = await platformModule.getAppDir(current);
        const installed = checkInstalled(appDir);
        if (installed) {
          console.log(`${AnsiEscapes.YELLOW}No platform specified, defaulting to "${current}".${AnsiEscapes.RESET}`);
          platform = current;
          break;
        }
      } catch (e) {}
    }

    if (!platform) {
      console.log(`${AnsiEscapes.RED}Could not find any installations of Discord.${AnsiEscapes.RESET}`);
      process.exit(process.argv.includes('--no-exit-codes') ? 0 : 1);
    }
```

asportnoy (changes requested):

asportnoy (approved):
yeet

PR #179, opened 2022-09-01
Move quick css to root

### LINKED ISSUE #153: Move quick CSS file to root folder
Makes more sense to have everything together, plus will make our lives easier in the future. Make sure to update migrations.

bop34:
Closes #153


asportnoy (changes requested):
The original migration should be updated to the new location, otherwise LGTM (tested and it works)

asportnoy (changes requested):
The issue from the last review isn't fixed yet, but the new changes look good to me besides that.

asportnoy (approved):

PR #188, opened 2022-09-05
npm ci

### LINKED ISSUE #186: Start using `npm ci`
Currently, Replugged uses `npm i` for installation of dependencies, which will break once someone pushes a different revision of the `package-lock.json`. `npm i` overwrites this file, however, `npm ci` does not and instead uses the file to install a reproducible set of dependencies.

Switching to `npm ci` will prevent alteration of `package-lock.json`, and will thus prevent merge conflicts.

asportnoy:
Closes #186

### DIFF
--- a/.github/workflows/eslint.yml
+++ b/.github/workflows/eslint.yml
@@ -23,7 +23,7 @@ jobs:
           node-version: 12
 
       - name: Install Node.js dependencies
-        run: npm i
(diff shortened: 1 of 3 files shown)


lexisother (approved):
Sidenote: still need to fix npm on UNIXes. The `fix-path` package doesn't help.

PR #189, opened 2022-09-08
Coremods setting folder

DracoClaw:
Moves coremods settings into their own folder. Also renames them to remove the `pc-` or `rp-` prefixes.

Additionally renames the settings for the community plugins to remove the `pc-` prefix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,9 @@ settings/*.json
 settings/quickcss/*
 !settings/quickcss/.gitkeep
 
+settings/coremods/*
+!settings/coremods/.gitkeep
+
 config.json
 *injected.txt
 *.nogit.*
--- a/.gitignore
+++ b/.gitignore
@@ -17,11 +17,8 @@ themes/*
 
 settings/*.json
 
-settings/quickcss/*
-!settings/quickcss/.gitkeep
-
-settings/coremods/*
-!settings/coremods/.gitkeep
+settings/**/*.json
+settings/quickcss/*.css
 
 config.json
 *injected.txt
--- a/src/Powercord/migrations/20220908-move-coremods-settings.js
+++ b/src/Powercord/migrations/20220908-move-coremods-settings.js
@@ -1,13 +1,13 @@
 const { existsSync, renameSync, unlinkSync } = require('fs');
+const { SETTINGS_FOLDER } = require('powercord/constants');
 const path = require('path');
 
 module.exports = () => {
-  const settingsPath = `${__dirname}/../../../settings`;
   const files = {
-    'pc-general.json': 'coremods/general.json',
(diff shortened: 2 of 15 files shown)


asportnoy (changes requested):
Nothing to request yet, just making sure this can be merged until I review.

lleyton (approved):
looks fine with me

.gitignore
```diff
@@ -20,6 +20,9 @@ settings/*.json
 settings/quickcss/*
 !settings/quickcss/.gitkeep
 
+settings/coremods/*
+!settings/coremods/.gitkeep
```
asportnoy, line 24:
```suggestion
settings/**/*.json
settings/quickcss/*.css
```
I think this is all you need. Make sure to remove the old line above this and make sure it works.

src/Powercord/migrations/20220908-move-coremods-settings.js
```diff
@@ -0,0 +1,59 @@
+const { existsSync, renameSync } = require('fs');
+
+module.exports = () => {
+  // Check if coremod settings exist.
+
+  if (existsSync(`${__dirname}/../../../settings/pc-general.json`)) {
```
asportnoy, line 6: Create variable with settings path so you don't have to repeat that

You could also make a helper function so you don't have to repeat as much. Or maybe have an object with all the paths and loop through that.

src/Powercord/migrations/index.js
```diff
@@ -1,5 +1,5 @@
 module.exports = async (powercord) => {
-  const settings = powercord.api.settings.buildCategoryObject('rp-migrations');
+  const settings = powercord.api.settings.buildCategoryObject('coremods/migrations');
```
asportnoy, line 2: This will not work the first time since the path won't have changed yet when this is running. I think that's fine, just update the migration to delete the existing file and they'll all run again.

asportnoy (changes requested):

src/Powercord/migrations/20220908-move-coremods-settings.js
```diff
@@ -0,0 +1,35 @@
+const { existsSync, renameSync, unlinkSync } = require('fs');
+const path = require('path');
+
+module.exports = () => {
+  const settingsPath = `${__dirname}/../../../settings`;
```
12944qwerty, line 5: Make this and coremods settings path a constant

12944qwerty (changes requested):

asportnoy (changes requested):
Let's call the folder core, not coremods

12944qwerty (approved):

DracoClaw:
Just include this in the re-write 😅

PR #190, opened 2022-09-09
Plugin embed uninstall

### LINKED ISSUE #183: [Feature Request] Uninstall plugin button on plugin embeds
sometimes you accidentally install a plugin and want to remove directly from the embed

also themes too

asportnoy:
Plugin embed button will show "uninstall" if plugin is installed. Will have same modal as the uninstall command.
Added a state for installing to give some user feedback that it's actually working.
Closes #183


src/Powercord/coremods/plugin-embeds/components/PluginEmbed.jsx
```diff
@@ -93,15 +92,42 @@ module.exports = function ({ match }) {
...
+              return;
+            }
             if (!data.isInstalled) {
-              cloneRepo(data.url, powercord, data.type).then(fetchInfo);
+              setData({
+                ...data,
+                isInstalling: true
+              });
+              // Give React time to update the button since the clone command causes the client to freeze for a sec
+              setTimeout(async () => {
```
lleyton, line 107: hmmm, I don't like this, I think it would be better to have a useEffect which runs the cloneRepo function once isInstalling is set to true

lleyton (changes requested):

lleyton (approved):
LGTM

PR #191, opened 2022-09-09
Editor env

asportnoy:
Unignore `.vscode`
Recommend VSCode extension, set it as the default formatter, and set format on save
Add `.editorconfig` with our indentation rules


lleyton (approved):
LGTM

PR #192, opened 2022-09-09
Fix plugin embed link

asportnoy:
Fix installer/gh links always showing up as just the github link if no data:
`https://replugged.dev/install?url=whatever/whatever` would show as `https://github.com/whatever/whatever`
`[some named link](https://github.com/whatever/whatever)` would show as `https://github.com/whatever/whatever`


asportnoy:
Need to merge #191 before this bc I based it on the wrong branch

src/Powercord/coremods/plugin-embeds/components/PluginEmbed.jsx
```diff
@@ -37,7 +37,7 @@ const {
   subHead
 } = getModule([ 'titleRegion' ], false);
 
-module.exports = function ({ match }) {
+module.exports = function ({ match, props }) {
```
lleyton, line 40: seems a bit strange to take a nested props object, why don't we just take in an href and title prop?

lleyton (changes requested):

src/Powercord/coremods/plugin-embeds/components/PluginEmbed.jsx
asportnoy: match is also a nested object. I don't personally see an issue with it.

lleyton: mmm, fair enough

lleyton (approved):
LGTM, but gonna wait on the cute cat to take a look

lexisother (approved):

lexisother:
Rebase time :trollface:

PR #195, opened 2022-09-11
Added /usr/bin/discord-canary to injector

Mobmaker55:
I hate git. /hj


asportnoy (changes requested):
Please add for the other platforms too (same format)

lleyton:
heh, I wonder if it would be worth it to try to locate the platform binary in the user's PATH

PR #201, opened 2022-09-22
Plaintext Patching

### LINKED ISSUE #181: Plaintext patcher
Let the fun begin.

lleyton:
Adds a new method (patchPlaintext) on the `powercord/webpack` API, which allows for directly patching module source code as plaintext.

Closes #181

### DIFF
--- a/src/fake_node_modules/powercord/util/debounce.js
+++ b/src/fake_node_modules/powercord/util/debounce.js
@@ -1,0 +1,11 @@
+// Credit goes to https://www.joshwcomeau.com/snippets/javascript/debounce/
+
+module.exports = (callback, wait) => {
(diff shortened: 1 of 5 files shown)


lexisother (changes requested):
Plugins load way after Webpack is done loading. This means that trying to patch an already required Webpack module will not work. (as it already exists in the require cache)
Instead of going at it in this manner, we should probably run our stuff in renderer.

asportnoy (commented):
Can you give an example of how this is used and what it can do that we couldn't do previously?

lexisother:
Patch unfetchable code, constants, local vars, etc.

asportnoy:
No longer relevant

PR #212, opened 2022-10-21
Fix all ESLint issues

lexisother:
Does as it says on the tin.


lleyton:
> merge 107 commits into [main](https://github.com/replugged-org/replugged)

uuuh, I don't think that's that you want, lexi-cat

lexisother:
silence, wench!

lleyton (approved):
Ready.

scripts/inject/index.ts
```diff
@@ -71,7 +71,7 @@ let platform: DiscordPlatform | undefined;
             platform = current;
             break;
           }
-        } catch (e) {}
+        } catch (_e) {}
```
asportnoy, line 74: Let's just remove the catch argument for all of these cases (not gonna comment on each one).
```suggestion
        } catch {}
```

src/globals.d.ts
```diff
@@ -10,9 +10,10 @@ declare global {
   };
 
   interface Window {
-    React: any;
+    React: typeof import("react");
```
asportnoy, line 13: Add import `import type React from "react";` above and change this to `React: typeof React;`

src/preload.ts
```diff
@@ -78,7 +78,8 @@ export type RepluggedNativeType = typeof RepluggedNative;
 contextBridge.exposeInMainWorld("RepluggedNative", RepluggedNative);
 
 const renderer = ipcRenderer.sendSync(RepluggedIpcChannels.GET_RENDERER_JS);
-webFrame.executeJavaScript(renderer);
+// eslint-disable-next-line @typescript-eslint/no-floating-promises
+webFrame.executeJavaScript(renderer).then((_r) => {});
```
asportnoy, line 82: Why do you need the promise here? Might be good to add a comment explaining.

src/types/addon.ts
```diff
@@ -10,6 +10,7 @@ export type Id = z.infer<typeof id>;
 
 export const author = z.object({
   name: z.string(),
+  // eslint-disable-next-line @typescript-eslint/naming-convention
   discord_id: z.string(),
```
asportnoy, line 14: Should we switch this parameter to camel case instead?

asportnoy (changes requested):
Pretty good for the most part, just some minor things here and there.

asportnoy (approved):
LGTM!

lleyton:
:leaSpin:

PR #214, opened 2022-10-23
Common modules

FifiTheBulldog:
Add common modules to renderer/modules/webpack. Just React for now, so that coremods and plugins can use it, but we can add more once this is in place.


lleyton (changes requested):
block merge

lleyton:
Superseded by #216

PR #217, opened 2022-10-24
Basic support for system electron packages

boehs:
Wouldn't it be cool if we didn't use f**king electron 13?

### DIFF
--- a/scripts/inject/platforms/linux.ts
+++ b/scripts/inject/platforms/linux.ts
@@ -14,6 +14,18 @@ const ProcessRegex = {
   dev: /discord-?development$/i,
 };
 
+const platformMappings: {[key in DiscordPlatform]: string} = {
+  stable: "discord",
+  dev: "discord-development",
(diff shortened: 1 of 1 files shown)


lleyton:
Hmmm, looks interesting. Maybe we should automatically patch in the `patchDiscordBin` function instead of prompting the user to do so?

boehs:
Funny you should say that. I actually did implement it this way initially, but decided against it at like 3am. The `findAppDir` function isn't currently aware of if the user is plugging or unplugging, so that would require adding a parameter to `findAppDir`, and adding file modifications to platform specific code (which isn't necessarily a bad thing, just something that hasn't been done before)

I will gladly change that though, I think the incredibly minimal effect on DX is much worth UX

boehs:
Forgot to mention: This should hopefully fix

- https://aur.archlinux.org/packages/discord_arch_electron
- https://aur.archlinux.org/packages/discord-canary-electron-bin

PR #218, opened 2022-10-24
Check function for plaintext patching

lleyton:


src/types/webpack.ts
```diff
@@ -14,10 +14,12 @@ export interface RegexReplacement {
 export type PlaintextReplacer = (source: string) => string;
 export interface PlaintextPatch {
   find?: string | RegExp;
+  check?: (source: string) => boolean;
```
lexisother, line 17: `source` can be optional here, since you can also use `check` for checking if a certain setting is enabled for example. If I'm correct `tsc` will complain about you not using `source` where it is required if you do not provide it.

lexisother (changes requested):

src/types/webpack.ts
lexisother: Noted.

lexisother (approved):

PR #219, opened 2022-10-30
Rename app.asar to work with e17 changes

FifiTheBulldog:
Electron 17 no longer looks for `app/` by default, preferring `app.asar` instead. This breaks our injection into the app. This PR modifies the injector script and `main/index.js` to rename Discord's `app.asar` to `app.orig.asar`, and the `app/` directory we create to `app.asar/`. This works with Electron 17 and is also backwards compatible with Electron 13.


lleyton (approved):
LGTM

PR #221, opened 2022-10-31
Timeout option for webpack `waitFor`

asportnoy:
Add timeout option to the webpack `waitFor` function. Will throw if nothing is returned before the specified duration.
Usage:
```js
webpack.waitForModule(YOUR_FILTER_HERE, { timeout: 1_000 });
```

The `raw` parameter has been moved to the options object. I tested that the timeout works as expected and the type definitions are still correct.


lleyton (approved):

lexisother (approved):
I love racing, they call me a raci

lleyton:
bad lexi-cat, bad girl

PR #222, opened 2022-11-05
Better managers

FifiTheBulldog:
Replaces the existing ignition implementation with a simpler version.

Features:
- No lifecycle targets. Plugins and coremods use separate entry points for loading plaintext patches (before webpack push) and regular patches (after webpack ready).
- No dependencies needed. These add a lot of complexity without a sizable use case.
- Coremods manager. Coremods load immediately before plugins at each stage.
- No more plugin class. We were only using it internally anyway, so we may as well replace it with the bare minimum.
- Load themes and quick CSS on startup alongside plugins.


Vendicated:
> No more plugin class. We were only using it internally anyway, so we may as well replace it with the bare minimum.

That's a terrible idea imo, you gain a lot from making plugin a class

ShadiestGoat:
> > No more plugin class. We were only using it internally anyway, so we may as well replace it with the bare minimum.
> 
> That's a terrible idea imo, you gain a lot from making plugin a class. things like this.logger, this.manifest, automatically tracked patches and more and the fact that `this` isn't scuffed in nested objects. It also ensures plugins are consistent, allows you to provide useful preimplemented methods and makes it possible to extend or refactor plugins in the future

I agree. @FifiTheBulldog, why do you want to remove the plugin class?

PR #223, opened 2022-11-05
Better injector

FifiTheBulldog:
Simplify the injector module a bit and make it more stable.

- WeakMap, Map, and Set are used to simplify the flow of adding, looking up, and removing injections
- Fewer custom types involved in the process, removed a lot of excess information that wasn't being used
- No more universal `uninject` function since injections don't have IDs anymore and are only available by reference--injections must be tracked by whatever created them (which is why `MiniInjector` exists)

### DIFF
--- a/src/renderer/modules/injector.ts
+++ b/src/renderer/modules/injector.ts
@@ -7,9 +7,13 @@ enum InjectionTypes {
   After,
 }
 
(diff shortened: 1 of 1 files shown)


lleyton (approved):
LGTM

PR #224, opened 2022-11-07
Loading Renderer as ESM

lleyton:
This change compiles the renderer as ESM, and loads it as such. This enables top-level await, allowing modules to await async code and block dependents, lazy webpack searchers for example.

The possibilities for this... are limitless. Much of the ignition code (maybe even the system itself) can be stripped, as we can be can be sure that async dependencies of a module are ready and exported *before dependents execute.* Additionally, this will help us avoid hacks such as `reactReady`, allowing for common modules to be as simple as:

```ts
export Prop = await waitForModule(...);
```


lexisother (approved):
Looks good as far as I can see.

PR #225, opened 2022-11-18
Allow injection script to not use switches

ShadiestGoat:
This is low priority PR only affects the injection script. 

All it does is removes the use of `process.argv` in favor of the use of a filtered array. This allows you to hardcode indecies like 2 or 3 without the need to worry about cmd flags & switches

### DIFF
--- a/scripts/inject/index.ts
+++ b/scripts/inject/index.ts
@@ -46,7 +46,7 @@ let platform: DiscordPlatform | undefined;
 
 (async () => {
   {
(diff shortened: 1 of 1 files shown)


ShadiestGoat:
*Renamed PR because it doesn't *fix* the script, it just *changes* it

asportnoy (changes requested):
Please run the linter and re-request my review.

ShadiestGoat:
Alright seems to to be linted & prettier-ed

asportnoy (approved):

PR #246, opened 2022-12-14
Support plugin css

asportnoy:
Plugins can now use CSS files. ESBuild automatically exports them with the same name as the renderer, so they can just be imported similar to themes.


12944qwerty (approved):

FifiTheBulldog (approved):
LGTM

PR #273, opened 2022-12-18
Expose paths in RepluggedNative (closes #272)

### LINKED ISSUE #272: [Feature Request] Expose plugin directory path
**Describe the feature request**
Give me a way to get the plugin directory programmatically.

**What Plugin would it be for?** 
NO plugin. i just want a function in the replugged/RepluggedNative constant to get the current plugin directory

**Expected behavior**
Output should be a string that contains the path to the plugin directory

```js
RepluggedNative.plugins.path: () => "/home/sammy/replugged/plugins"
```

SammCheese:
Adds 

```js
RepluggedNative.themes.path()
RepluggedNative.quickCSS.path()
RepluggedNative.plugins.path()
RepluggedNative.settings.path()
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "replugged",
-  "version": "4.0.0-beta0.18",
+  "version": "4.0.0-beta0.19",
   "description": "A lightweight @discord client mod focused on simplicity and performance",
   "license": "MIT",
   "main": "dist/main/index.js",


asportnoy (changes requested):
- You don't need to go through IPC for this. Better to just import the path from util and append the path for each thing.
- The path will not change, so we can make them just strings instead of functions
- Let's move them to `replugged` instead of `RepluggedNative`

SammCheese:
yeah no, it has to be done in main, which only ipc does

> * You don't need to go through IPC for this. Better to just import the path from util and append the path for each thing.
> * The path will not change, so we can make them just strings instead of functions
> * Let's move them to `replugged` instead of `RepluggedNative`

PR #275, opened 2022-12-18
Allow setting webpack module exports (closes #271)

### LINKED ISSUE #271: Function injection errors/doesn't callback
**Describe the bug**
When attempting to inject into a function, the first time it fails with an error, every other time it doesn't error but doesn't do anything (doesn't seem to actually be injecting/callback not getting called).

**To Reproduce**
Example code for reproduction:
```ts
import { AnyFunction, Injector, webpack } from "replugged";
const inject = new Injector();

function moduleFindFailed(moduleName: string): void {
  console.error(`Failed to find ${moduleName} module! Cannot continue`);
}

export async function start(): Promise<void> {
  const tagRenderMod = await webpack.waitForModule<{ x: AnyFunction }>(
    webpack.filters.bySource(".botTagCompact"),
  );
  if (!tagRenderMod) return moduleFindFailed("tagRenderMod");

  inject.instead(tagRenderMod, "x", (args, fn) => {
    console.log(`#########################`, args, fn);
    return fn(...args);
  });
}

export function stop(): void {
  inject.uninjectAll();
}
```

**Expected behavior**
Everytime, the function should be injected without errors, and it the callback should be called.

**Desktop**

- OS: Window 11 (22H2)
- Replugged version: [v4.0.0-beta0.17](https://github.com/replugged-org/replugged/releases/tag/v4.0.0-beta0.17)

**Additional context**

- [X] I checked if Discord and/or Replugged are up to date

[Conversation in discord with asportnoy](https://canary.discord.com/channels/1000926524452647132/1000955966520557689/1053887327602216973)

Penguin-Spy:
By default, most (if not all) webpack modules define their exports using a function that only provides getters. This means that the Replugged injector cannot replace the function with it's injected version, and so injection fails.  

This PR replaces that function with one that behaves identically, but creates a setter for the exports as well. I have tested this with the example code provided in #271 and it properly injects into the method.

I also added an error handler for the plugin load function because otherwise the errors are dropped silently and it's really annoying to figure out why something won't load.


12944qwerty (approved):

asportnoy (approved):
Nice!

PR #278, opened 2022-12-19
Don't bundle React

FifiTheBulldog:
A bunch of goodies centered around React:

- Remove the React dependency, so that it will no longer be compiled with our renderer code. We have `@types/react` for types, and that is more than enough.
- Customize our JSX factory.
- Don't export types from replugged.ts, so that Zod also isn't in our renderer code when compiled.
- Split common modules into several files, and hopefully maximize the stability and maintainability of this part of the mod
- Expose the logger module on window.replugged.
- Expose common modules on replugged.common, rather than replugged.webpack.common.
- Add proper types for hljs.
- Delete types/webpack-common.ts, and move those types to where they are actually needed in common modules. Avoid duplicating stuff.

Some of these are breaking changes for those who have made plugins already, but it should be early enough for plugin devs to adapt.


src/renderer/coremods/settings/index.tsx
```diff
@@ -12,13 +12,15 @@ export default class SettingsMod extends Coremod {
 
   start() {
     settingsTools.addAfter("Billing", [
+      /* eslint-disable */
```
asportnoy, line 15: Why?

src/renderer/coremods/settings/lib.tsx
```diff
@@ -39,6 +40,7 @@ export const settingsTools: SettingsTools = {
   rpSections: [],
   rpSectionsAfter: {},
   addSection(opts) {
+    // eslint-disable-next-line
```
asportnoy, line 43: Why?

src/renderer/modules/webpack/common/index.ts
```diff
@@ -0,0 +1,63 @@
...
+): void {
+  const timeout = setTimeout(
+    () => error("Replugged", "CommonModules", void 0, `Could not find module "${name}"`),
+    5_000,
+  );
+  void moduleImport.then((mod) => {
+    clearTimeout(timeout);
+    cb(mod);
+  });
+}
```
asportnoy, line 17: I think we should actually stop the queries, not just do this

asportnoy (changes requested):

SammCheese (commented):
Looks good to me

src/renderer/coremods/settings/index.tsx
FifiTheBulldog: Because the linter hates when the function name is capitalized but it's not used as a class. Hopefully temporary, but this way eslint doesn't bother me while I'm focusing on more important stuff.

src/renderer/coremods/settings/lib.tsx
FifiTheBulldog: Same as in index.tsx

src/renderer/modules/webpack/common/index.ts
FifiTheBulldog: My objection to stopping the query is mostly theoretical given the speed of my computer: if for some reason a module takes longer than five seconds to be found, but eventually becomes available, then stopping the query will make the erroneous assertion that the filter is wrong and will prevent there from being any information to the contrary (i.e. the presence of the module in `replugged.common` despite the error message). Besides, the performance impact of this for the limited number of common modules is negligible at worst.

PR #284, opened 2022-12-21
Restructure coremods to match plugins

FifiTheBulldog:
Switches coremods to follow the same pattern as plugins (minus the manifest), including:

- No more `Coremod` class
- Export start/stop as functions
- Plaintext patches belong in a separate file

Also switched the noDevtoolsWarning coremod to use `DiscordNative.window.setDevtoolsCallbacks()` like Powercord and old Replugged did instead of a plaintext patch, just for easier maintenance unless we decide to bring Replugged to web browsers where DiscordNative doesn't exist.


src/renderer/coremods/noDevtoolsWarning/index.ts
```diff
@@ -1,32 +1,17 @@
...
+{
+  find: "setDevtoolsCallbacks",
+  replacements: [
+    {
+      match: /if\(.{0,10}\|\|"0.0.0"!==.{0,2}\.remoteApp\.getVersion\(\)\)/,
+      replace: "if(false)"
+    }
+  ]
+}
+*/
```
asportnoy, line 13: It's not good practice to comment out code like this. That's what Git history is for.
```suggestion
```

asportnoy (changes requested):

src/renderer/coremods/noDevtoolsWarning/index.ts
FifiTheBulldog: Fixed

asportnoy (approved):

PR #285, opened 2022-12-21
Better settings placeholder

asportnoy:
Some people didn't understand that the settings page was a placeholder so I made it more clear. Then I decided the page was a bit boring so I ✨ spiced it up a bit ✨.

### DIFF
--- a/src/renderer/coremods/settings/pages/General.tsx
+++ b/src/renderer/coremods/settings/pages/General.tsx
@@ -1,4 +1,4 @@
-import { React } from "src/renderer/modules/webpack/common";
+import React from "../../../modules/webpack/common/react";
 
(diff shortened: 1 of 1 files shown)


src/renderer/coremods/settings/pages/General.tsx
```diff
@@ -1,3 +1,21 @@
+import { React } from "src/renderer/modules/webpack/common";
```
FifiTheBulldog, line 1: Let's import React directly from the file where we find it:
```ts
import React from "src/renderer/modules/webpack/common/react";
```

FifiTheBulldog (changes requested):

FifiTheBulldog (approved):

PR #287, opened 2022-12-22
Badges

asportnoy:
There are a few things that will need to be moved to common modules/components once we have the framework, but I think this is good to release in its current form.


FifiTheBulldog (changes requested):
There's an alignment issue in the user popout when badges need more than one line:

Screenshot 2022-12-22 at 12 35 16 AM
Screenshot 2022-12-22 at 12 35 30 AM

Other than that, looks good.

FifiTheBulldog (approved):

PR #293, opened 2022-12-22
Nuke sentry into oblivion (closes #250) + minor preload changes

### LINKED ISSUE #250: No Track Coremod

SammCheese:
Prevents sentry from being fetched ( closes #250 )


SammCheese:
@FifiTheBulldog @asportnoy need a review here

SammCheese:
also removes the preload console.log. its not needed.

package.json
```diff
@@ -53,5 +53,8 @@
   },
   "files": [
     "dist"
-  ]
+  ],
+  "dependencies": {
+    "node-fetch": "^3.3.0"
+  }
```
asportnoy, line 59: Don't need to add it, it's a global function.

src/main/index.ts
```diff
@@ -100,6 +101,36 @@ electron.protocol.registerSchemesAsPrivileged([
...
+electron.app.once("ready", (_) => {
+  let sentryBlocked = false;
+
+  ipcMain.on(RepluggedIpcChannels.REGISTER_RELOAD, () => {
+    console.log("sentryblock reset");
+    sentryBlocked = false;
+  });
+
+  session.defaultSession.webRequest.onBeforeRequest(async (details, callback) => {
+    if (!sentryBlocked) {
```
asportnoy, line 113: This seems like an unnecessarily janky way to do this. For reference, here's how Powercord did it: https://github.com/replugged-org/replugged/blob/legacy/src/Powercord/coremods/no-track/index.js

src/main/index.ts
```diff
@@ -100,6 +101,36 @@ electron.protocol.registerSchemesAsPrivileged([
   },
 ]);
 
+electron.app.once("ready", (_) => {
```
asportnoy, line 104: We already have once ready in this file, why do we need another?

src/preload.ts
```diff
@@ -82,9 +82,14 @@ contextBridge.exposeInMainWorld("RepluggedNative", RepluggedNative);
...
 
-// Get and execute Discord preload
-// If Discord ever sandboxes its preload, we'll have to eval the preload contents directly
-const preload = ipcRenderer.sendSync(RepluggedIpcChannels.GET_DISCORD_PRELOAD);
-if (preload) {
-  require(preload);
+try {
+  window.addEventListener("beforeunload", () =>
+    ipcRenderer.send(RepluggedIpcChannels.REGISTER_RELOAD),
+  );
```
asportnoy, line 88: Out of scope?

asportnoy (changes requested):

package.json
SammCheese: the "global function" does not exist at the time the code runs.

src/main/index.ts
SammCheese: Seperation. Its easier to read separated callbacks instead of something mushed together.

src/main/index.ts
CanadaHonk: This method also adds delay and overhead.

package.json
thewilloftheshadow: It's only global in node v18 right? Replugged goes down to node v14

SammCheese:
bruh this shit formatted the whole file

SammCheese:
Good to go now

src/main/index.ts
```diff
@@ -1,7 +1,7 @@
 import { dirname, join } from "path";
 
-import electron from "electron";
-import type { RepluggedWebContents } from "../types";
+import electron, { ipcMain, session } from "electron";
+import { RepluggedIpcChannels, RepluggedWebContents } from "../types";
```
asportnoy, line 4: Should be reverted since no longer needed

asportnoy (approved):
much better

PR #298, opened 2022-12-24
Updater

### LINKED ISSUE #239: Updater Coremod
Uses asar and GitHub releases

asportnoy:
Implements the plugin/theme updater! To use it, run:
```js
await replugged.updater.checkAllUpdates()
await replugged.updater.installAllUpdates()
```
Defaults to limiting checks to every 15 minutes. No auto updating yet. Note that the GitHub API has a limit of 60 requests per hour when unauthenticated, so be careful with how often you run it.

Will need updates in the future to work with the UI (once it exists) and work with the store (once it also exists). Some of the code could probably be shared with the plugin/theme installer (once it exists too), so that's cool.

Closes #239


src/main/ipc/updater.ts
```diff
@@ -0,0 +1,161 @@
...
+    });
+  } catch (err) {
+    return {
+      success: false,
+      // @ts-expect-error err tbd
+      error: err,
+    };
+  }
+
+  const asset = res.data.assets.find((asset) => asset.name.endsWith(".asar"));
```
E-boi, line 63: Instead of looking for if a file has .asar it should look for the filename because ideally the filename will be plugin/theme id. Also this makes it so monorepos wont work like [mine](https://github.com/E-boi/replugged-plugins).

asportnoy:
Pointing to `better-settings` for now for review purposes, but will not merge until that PR is merged

src/main/ipc/updater.ts
```diff
@@ -0,0 +1,162 @@
...
+import fetch from "node-fetch";
+import { join } from "path";
+import { Hash, createHash } from "crypto";
+
+const octokit = new Octokit();
+
+async function getHashRecursive(path: string, hash: Hash): Promise<void> {
+  const files = await readdir(path, { withFileTypes: true });
+
+  for (const file of files) {
```
asportnoy, line 22: todo: make this concurrent

src/renderer/managers/updater.ts
```diff
@@ -0,0 +1,218 @@
...
+  } = entity;
+  if (!updater) {
+    error("Replugged", "Updater", void 0, `Entity ${id} has no updater info`);
+    return;
+  }
+
+  const res = await window.RepluggedNative.updater.check(updater.type, updater.id);
+
+  if (!res.success) {
+    error("Replugged", "Updater", void 0, `Update check failed: ${res.error}`);
```
asportnoy, line 79: todo: warn

src/renderer/managers/updater.ts
```diff
@@ -0,0 +1,218 @@
...
+  ) {
+    if (verbose) log("Replugged", "Updater", void 0, `Entity ${id} was checked recently, skipping`);
+    return;
+  }
+
+  const {
+    manifest: { updater },
+  } = entity;
+  if (!updater) {
+    error("Replugged", "Updater", void 0, `Entity ${id} has no updater info`);
```
asportnoy, line 72: todo: warn

src/renderer/managers/updater.ts
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  const newHash = await window.RepluggedNative.updater.getHash(entity.manifest.type, entity.path);
+
+  // update settings
+  updaterState.set(id, {
+    ...updateSettings,
+    available: false,
+    hash: newHash,
+  });
```
asportnoy, line 180: todo: do before reloading

FifiTheBulldog (commented):
I'm very hesitant to approve this at this stage for a few reasons:

- So much of it is going to need to be changed for the store.
- I'm not sure how reliable this is; in my testing, this appears to install updates, but it fails to reload them, throwing bizarre syntax errors instead. (There's a systemic problem here that renders the reload function useless anyway as long as plugins are shipped as asars.)
- I'm working on a major update for the plugin manager that will probably include some breaking changes for this.

src/renderer/managers/plugins.ts
```diff
@@ -172,14 +172,14 @@ export async function list(): Promise<RepluggedPlugin[]> {
  * Some plugins may require Discord to be reloaded to apply changes.
  */
 export async function reload(id: string): Promise<void> {
-  const plugin = plugins.get(id);
+  const plugin = plugins.get(id) || Array.from(plugins.values()).find((x) => x.path === id);
```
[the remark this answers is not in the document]
asportnoy: Adds consistency with the theme manager since that requires `.asar` for asars

asportnoy:
@FifiTheBulldog updated to work with the new plugin manager. GitHub will be an update source on top of the store. I designed it so it can be extended for the store once it's out.

FifiTheBulldog (approved):
Merge whenever you're ready

PR #299, opened 2022-12-24
Better settings

FifiTheBulldog:
Breaking changes for settings. Instead of fully asynchronous, direct access to settings over IPC, this is a fresh implementation:

- Uses the same RepluggedNative.settings machinery to communicate with the file system over IPC
- Rename NamespacedSettings to SettingsManager, and replugged.settings.get to replugged.settings.manager
- Stores a copy of the settings data for each namespace in memory in the window
- All methods to modify settings data are now synchronous
- Updates are queued in a map on each SettingsManager instance, and then once a namespace's settings are no longer being actively modified, updates are performed asynchronously with a delay enforced by setTimeout.

This also gets rid of the API and EntityBase classes, as they are not needed anymore. If APIs need EventTarget functionality, they can extend EventTarget directly, like notices and commands now do.


src/renderer/apis/settings.ts
```diff
@@ -1,64 +1,169 @@
...
+ * export function start() {
+ *   cfg.set('hello', 'world');
+ *   console.log(cfg.get('hello')); // world
+ * }
+ * ```
+ * @typeParam T Type definition for the settings to manage in the namespace.
+ * This will be an object with strings as keys, and JSON-serializable values.
+ * @param namespace Namespace to manage. A namespace is an ID (for example, the ID of a plugin) that uniquely identifies it.
+ * All settings are grouped into namespaces.
+ * Settings for a namespace are stored in `replugged/settings/NAMESPACE.json`.
```
asportnoy, line 156: Clarify where it is stored
```suggestion
 * Settings for a namespace are stored in `settings/NAMESPACE.json` within the [Replugged data folder](https://docs.replugged.dev/#installing-plugins-and-themes).
```

asportnoy (commented):
Minor doc clarification

asportnoy (approved):

PR #309, opened 2022-12-31
Lodash types

FifiTheBulldog:
Add types for lodash as `window._`, and add lodash to common modules.

### DIFF
--- a/src/globals.d.ts
+++ b/src/globals.d.ts
@@ -3,7 +3,7 @@
 import { WebpackChunkGlobal } from "./types/discord";
 import * as replugged from "./renderer/replugged";
 import { RepluggedNativeType } from "./preload";
-import type React from "react";
+import type Lodash from "lodash";
 
 declare global {
   export var appSettings: {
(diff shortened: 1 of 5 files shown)


src/globals.d.ts
```diff
@@ -3,26 +3,31 @@
...
   };
 
   interface Window {
-    React: typeof React;
     RepluggedNative: RepluggedNativeType;
+    DiscordNative: typeof DiscordNative;
     replugged: typeof replugged;
     // eslint-disable-next-line @typescript-eslint/naming-convention
     webpackChunkdiscord_app: WebpackChunkGlobal;
+    _: typeof _;
```
asportnoy, line 19: Shouldn't this be in common modules

FifiTheBulldog: It is, but lodash is available globally and I see no harm in allowing devs to use that instead of an inconvenient import.

asportnoy: I think we should do the same with React then. That way ppl can import it directly.

FifiTheBulldog: Hmm, maybe, although `window._` is created by Discord while `window.React` would be us polluting the window...out of scope for this though

asportnoy: I think it's worth it but we can create a separate PR for that.

asportnoy (approved):

PR #310, opened 2023-01-03
Plugin exports and typedef stuff

FifiTheBulldog:
Step 1 of improving the plugin manager: add the ability to access the full exports of a plugin from `replugged.plugins`, so that plaintext patches can easily access exported functions and properties from plugins if needed.

Also a bunch of typedef stuff, including adding type-fest as a dependency to get some juicy utility types and stop defining them ourselves, as well as an attempt to improve the typedef for WebpackChunk.push. Maybe we should spin some of these off into separate packages.

I'll leave further plugin manager changes for a separate PR.

Merge conflicts, here I come


cspell.json
```diff
@@ -4,44 +4,36 @@
   "dictionaryDefinitions": [],
   "dictionaries": [],
   "words": [
-    // replugged terminology
```
asportnoy, line 7: Any reason why you removes these?

src/renderer/apis/settings.ts
```diff
@@ -18,7 +20,7 @@ type SettingsUpdate<T> =
  * Once the settings data has been copied into the `SettingsManager`, it can be read and written synchronously.
  * The `SettingsManager` automatically queues and dispatches updates to the file system in the background.
  */
-export class SettingsManager<T extends Record<string, unknown>> {
+export class SettingsManager<T extends Record<string, Jsonifiable>> {
```
asportnoy, line 23: Can't you do away with the record and make the whole thing JSON?

src/renderer/apis/settings.ts
```diff
@@ -156,7 +158,7 @@ const managers = new Map<string, unknown>();
  * Settings for a namespace are stored in `settings/NAMESPACE.json` within the [Replugged data folder](https://docs.replugged.dev/#installing-plugins-and-themes).
  * @returns Manager for the namespace.
  */
-export async function init<T extends Record<string, unknown>>(
+export async function init<T extends Record<string, Jsonifiable>>(
```
asportnoy, line 161: Same

src/renderer/managers/plugins.ts
```diff
@@ -1,83 +1,40 @@
...
-          );
-        }
-      },
-    });
-    plugins.set(plugin.manifest.id, pluginWrapper);
-  } catch (e: unknown) {
-    error("Plugin", plugin.manifest.id, void 0, "Plugin failed to load\n", e);
-  }
+export function getExports(id: string): PluginExports | undefined {
+  return plugins.get(id)!.exports;
```
asportnoy, line 30:
```suggestion
  return plugins.get(id)?.exports;
```

asportnoy (changes requested):

src/renderer/apis/settings.ts
FifiTheBulldog: No, because settings are read/written like values in a map. The settings API imitates the methods of `Map`, but the type is defined using objects to allow different types for different keys. The top level of settings for a thing will always be a Record<string,Jsonifiable>, never an array or any other JSON value.

src/renderer/apis/settings.ts
FifiTheBulldog: See other one

src/renderer/managers/plugins.ts
FifiTheBulldog: Plugin exports can be undefined if a plugin only uses plaintext patches and has no renderer script. I would prefer for this to throw an error if the plugin isn't even loaded, so the behavior clearly distinguishes between situations.

asportnoy: In that case, explicitly check for undefined and throw a descriptive message. Non-null assertions should only be used if you know 100% that something is non-null, which is not the case here.

cspell.json
FifiTheBulldog: I didn't do that myself; I think either the Prettier or eslint plugin in VSCode did that automatically when I edited that file, because it sees comments in JSON and labels them (correctly) as invalid.

asportnoy: Interesting

src/renderer/managers/plugins.ts
```diff
@@ -1,83 +1,44 @@
...
-    });
-    plugins.set(plugin.manifest.id, pluginWrapper);
-  } catch (e: unknown) {
-    error("Plugin", plugin.manifest.id, void 0, "Plugin failed to load\n", e);
+export function getExports(id: string): PluginExports | undefined {
+  const plugin = plugins.get(id);
+  if (plugin) {
+    return plugin.exports;
   }
+  throw new Error(`Plugin "${id}" does not exist or is not loaded`);
```
asportnoy, line 34: IMO it's better to handle the edge cases first and then return the value at the end

asportnoy (approved):

PR #294, opened 2022-12-22
More modules

12944qwerty:
Currently includes:
 
- [x] Category Component
- [x] FormItem, FormText, and Divider Components (all for settings)
- [x] SwitchItem Component
- [x] ContextMenu Module
- [x] ContextMenu Components (not all of the components are included yet)
- [x] Modal Components and Module
- [x] Type FluxDispatcher
- [x] Tooltip Component
- [x] Button and ButtonItem Components

PR #306, opened 2022-12-28
Fix Module Stuff

12944qwerty:

PR #312, opened 2023-01-04
Revert More modules

asportnoy:
This reverts commit #294. The PR was not ready and should not have been merged in the first place. The changes for it will not be ready for a bit due to @12944qwerty's availability, so we'll revert this for now until that's done.


FifiTheBulldog (commented):
I'm happy to pick up things that are unfinished and contribute to #306 so that the finished version of more-modules makes it into main, but we should not revert the changes that have already been made; that'll make even more of a mess down the road.

SammCheese:
I agree with Fifi here. The Changes made should stay in case some plugins already use the components.

12944qwerty:
Other commits on the main branch have already modified some files from this (like plugin exports and typedef stuff). This would break that stuff.

PR #316, opened 2023-01-07
Fix SwitchItemTypes "onChange" parameter

SammCheese:
become as gods


SammCheese:
@FifiTheBulldog @asportnoy because cant add reviewers

asportnoy (approved):

PR #318, opened 2023-01-08
Use plugin/theme path in updater

E-boi:
use the path from plugin/theme instead of looking for a .asar from a release.

### DIFF
--- a/src/main/ipc/updater.ts
+++ b/src/main/ipc/updater.ts
@@ -42,7 +42,7 @@ async function getFolderHash(path: string): Promise<string> {
 
 async function github(
   identifier: string,
-  path: string,
+  id: string,
 ): Promise<UpdateCheckResultSuccess | UpdateCheckResultFailure> {
   const [owner, repo] = identifier.split("/");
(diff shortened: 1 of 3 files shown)


src/main/ipc/updater.ts
```diff
@@ -66,7 +67,8 @@ async function github(
     };
   }
 
-  const asset = res.data.assets.find((asset) => asset.name.endsWith(".asar"));
+  const asset = res.data.assets.find((asset) => asset.name === path);
```
asportnoy, line 70: I think you should use the ID instead since the path could change.

asportnoy (changes requested):

src/main/ipc/updater.ts
E-boi: wdym ID? the path is just the id with .asar added at the end

asportnoy: The path is the file name, which doesn't necessarily have to be the ID.

asportnoy (approved):

PR #321, opened 2023-01-10
Disable plugins

### LINKED ISSUE #307: Disable/enable plugins
Should work the same as themes

asportnoy:
Adds disabling/enabling plugins. Did some slight refactoring and tweaked the themes manager for better consistency. Closes #307


asportnoy:
Will merge this with the plugin settings UI

PR #330, opened 2023-01-11
feat/fix: Add `forceUpdateElement` utility method and resolve reload crash

### LINKED ISSUE #291: Fix crash on reload
When reloading discord (ctrl+r), the console indicates that Discord crashed (`AppCrashedFatalReport lastCrash: `) with an object including the current date). I believe this is also why you're brought back to the friends view after reloading. We need to figure out how to properly handle a reload. Maybe there's something we need to stop to allow it to successfully exit?

### LINKED ISSUE #315: Add `forceUpdateElement` to utils
You can take the code from my Mention Cache Fix plugin:
https://github.com/asportnoy/mention-cache-fix/blob/69592d0d9c94b6675660cf48a5caaf928363306d/src/index.ts#L81-L92

Just needs some doc comments

GriefMoDz:
### Proposed Changes:
- Add `forceUpdateElement` utility method (closes #315)
    - I've added `@types/react-reconciler` as a dependency to make use of an actual type for `Fiber`. 
- Resolve crash on reload (closes #291)
    - Enforcing `bypassCSP` on a registered privileged scheme is the cause of this.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -61,7 +61,10 @@ class BrowserWindow extends electron.BrowserWindow {
   }
 }
 
-Object.defineProperty(BrowserWindow, "name", { value: "BrowserWindow", configurable: true });
+Object.defineProperty(BrowserWindow, "name", {
+  value: "BrowserWindow",
+  configurable: true,
+});
 
 const electronExports: typeof electron = new Proxy(electron, {
   get(target, prop) {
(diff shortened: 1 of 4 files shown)


src/main/index.ts
```diff
@@ -93,7 +96,7 @@ electron.protocol.registerSchemesAsPrivileged([
   {
     scheme: "replugged",
     privileges: {
-      bypassCSP: true,
+      // bypassCSP: true,
```
asportnoy, line 99:
```suggestion
```

asportnoy (approved):
Can confirm fix works. Nice find!

PR #335, opened 2023-01-15
Badges coremod improvements

asportnoy:
- Create new `gotoOrJoinServer` util
- Add click events for more replugged badges, including joining the server
- Use common `Tooltip` component

### DIFF
--- a/src/renderer/coremods/badges/badge.tsx
+++ b/src/renderer/coremods/badges/badge.tsx
@@ -1,20 +1,9 @@
-import {
-  filters,
-  getByProps,
-  getBySource,
-  getFunctionBySource,
-  waitForModule,
-} from "../../modules/webpack";
-import React from "@common/react";
+import { getByProps, getBySource } from "../../modules/webpack";
+import { React } from "@common";
 import "./badge.css";
 import Badges from "./badges";
-
-type Tooltip = React.FC<{
-  text?: string;
-  position?: "top" | "bottom" | "left" | "right";
-  spacing?: number;
-  children: (props: React.HTMLAttributes<HTMLSpanElement>) => React.ReactElement;
-}>;
+import { Tooltip } from "@components";
+import { gotoOrJoinServer } from "../../util";
 
 type Clickable = React.FC<
   Record<string, unknown> & {
@@ -35,17 +24,15 @@ interface BadgeProps {
   onClick?: () => void;
 }
 
-export const getBadgeComponent = async (): Promise<(args: BadgeProps) => React.ReactElement> => {
-  const tooltipMod = await waitForModule<Record<string, Tooltip>>(
-    filters.bySource(/shouldShowTooltip:!1/),
-  );
-
-  // todo: common component
-  const Tooltip = tooltipMod && getFunctionBySource<Tooltip>(/shouldShowTooltip:!1/, tooltipMod);
-  if (!Tooltip) {
-    throw new Error("Failed to find Tooltip component");
-  }
-
+export const Base = ({
+  color,
+  tooltip,
+  tooltipPosition,
+  className,
+  children,
+  gap,
+  onClick,
+}: BadgeProps): React.ReactElement => {
   // todo: common component
   const Clickable = getBySource("renderNonInteractive") as Clickable | undefined;
   if (!Clickable) {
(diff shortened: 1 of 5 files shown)


src/renderer/coremods/badges/badge.tsx
```diff
@@ -35,21 +24,25 @@ interface BadgeProps {
...
+export const Base = ({
+  color,
+  tooltip,
+  tooltipPosition,
+  className,
+  children,
+  gap,
+  onClick,
+}: BadgeProps): React.ReactElement => {
   // todo: common component
```
12944qwerty, line 36: Just make it a common component rn lol

src/renderer/coremods/badges/badge.tsx
```diff
@@ -71,108 +64,137 @@ export const getBadgeComponent = async (): Promise<(args: BadgeProps) => React.R
...
-      <Clickable className={clickableClass.clickable} onClick={onClick || (() => undefined)}>
+  const child = (
+    <div
+      className={`${profileBadge22} replugged-badge ${className || ""}`}
+      style={{ color: `#${color || "7289da"}` }}>
+      {children}
+    </div>
+  );
+  return (
+    <Clickable className={clickableClass.clickable} onClick={onClick || (() => undefined)}>
```
12944qwerty, line 75: If onClick isn't defined, then the badge shouldn't be clickable

src/renderer/coremods/badges/badge.tsx
```diff
@@ -71,108 +64,137 @@ export const getBadgeComponent = async (): Promise<(args: BadgeProps) => React.R
...
-            </span>
-          )}
+          {child}
         </Tooltip>
-      </Clickable>
-    );
-  };
+      ) : (
+        child
+      )}
```
12944qwerty, line 85: There is a `shouldShow` prop that does this for you lol

```jsx
<Tooltip
  shouldShow={Boolean(tooltip)}
  text={tooltip!}
  position={tooltipPosition || "top"}
  spacing={gap === false ? 0 : 12}>
  {child}
</Tooltip>
```

12944qwerty (changes requested):

src/renderer/util.ts
```diff
@@ -75,3 +78,87 @@ export function forceUpdateElement(selector: string, all = false): void {
 
   elements.forEach((element) => getOwnerInstance(element)?.forceUpdate());
 }
+
+type Invite = Record<string, unknown> & {
```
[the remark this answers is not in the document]
asportnoy: It's not in `discord-types`. This is only a partial type for this one specific usage, so unless there's use for it elsewhere, I think we should leave it here.

src/renderer/coremods/badges/badge.tsx
asportnoy: The clickable class gives it the cursor, which I think it should have due to the tooltip.

src/renderer/coremods/badges/badge.tsx
asportnoy: I think the current way of doing it is better. No reason to include extra elements if they're not being used.

12944qwerty: It's not an extra element... it's just a prop that's already added. You';d just have to replace those lines with the codeblock I sent

asportnoy: If the element isn't doing anything then it is an extra element.

src/renderer/coremods/badges/badge.tsx
12944qwerty: I meant that if onclick isn't defined, then the cursor shouldn't be the clickable type of cursor lol. If it doesn't do anything, there shouldn't be any need to have the clickable cursor.

src/renderer/util.ts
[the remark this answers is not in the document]
12944qwerty: It's still a discord type.

src/renderer/coremods/badges/badge.tsx
asportnoy: All of the Discord badges show a pointer even if they don't do anything onclick.

src/renderer/coremods/badges/badge.tsx
12944qwerty: You're literally reducing lines into one prop line?

12944qwerty (approved):
for a sec I was like "to test join server, I need to leave and then only I can join... hmmm"

PR #347, opened 2023-01-19
Remove plugin preload

asportnoy:
Remove plugin preload and main manifest keys since those will not be added.

Also removes console errors for plugin CSS files not existing.


colin273 (changes requested):
I don't think the error removal quite worked.

Also, while we're at it, could we get rid of the dependencies and dependents fields? Those aren't being used either.

colin273 (approved):

PR #348, opened 2023-01-20
Plugin settings UI

### LINKED ISSUE #327: Plugin Settings Pages

asportnoy:
Adds setting page functionality for Replugged. You can see an example of this being implemented in my [role dividers plugin](https://github.com/asportnoy/role-dividers/pull/6/files)

- Add `Settings` export to the plugin renderer (element for the settings)
- Add `defaultSettings` option to the settings manager to make it easier to have default settings without having to specify the fallback every time.
  - Specifying a `fallback` will override this value.
  - I added another type property for specifying which keys have default values. Unfortunately, I couldn't infer this from the property itself, so it'll either have to specified or users will have to non-null assert/check themselves.
- Add a custom `useSetting` React hook to the settings manager. This can be used to easily apply the `value` and `onChange` properties to the inputs.
- Add `ErrorBoundary` component
- Fix type on `ModalCloseButton`
- Rename `SwitchItem` to `Switch` and update types accordingly (BREAKING CHANGE)
- Add back `SwitchItem` with selector that was used before. (BREAKING CHANGE)
- Rename `Input` to `TextInput` (BREAKING CHANGE)
- Fix misc component selectors

Closes #327


src/renderer/coremods/settings/pages/Addons.tsx
```diff
@@ -33,6 +33,17 @@ function getManager(type: AddonType) {
   throw new Error("Invalid addon type");
 }
 
+function getSettingsElement(id: string, type: AddonType) {
+  if (type === AddonType.Plugin) {
+    return window.replugged.plugins.plugins.get(id)?.exports?.Settings;
```
12944qwerty, line 38: Shouldn't settings be gotten in a different way? Could it work through the manifest? I'm not sure how I feel about it being gotten from exports...
And otherwise, if you continue using this method, `replugged.plugins.getExports` exists (and shouldn't you import this somehow)

12944qwerty (changes requested):

src/renderer/coremods/settings/pages/Addons.tsx
asportnoy: @colin273 said I should do it this way, I originally had a manifest key. I'll update to getExports later.

12944qwerty (approved):

src/renderer/apis/settings.ts
```diff
@@ -1,4 +1,5 @@
 import type { Jsonifiable } from "type-fest";
+import { React } from "@common";
```
colin273, line 2: I don't think we should be involving webpack modules in our APIs. Webpack modules can certainly *use* the APIs, but they should not be *part of* the APIs.

colin273 (changes requested):
Works well, but I have a couple of concerns--see comments.

colin273 (approved):

PR #352, opened 2023-01-22
feat(mod): Add i18n support

### LINKED ISSUE #331: i18n
Set up the i18n module for internationalizing replugged. Sub in the hardcoded messages for i18n. Create new messages where needed (should be done in Weblate I think).

GriefMoDz:
### Proposed Changes:
* Added an `i18n` webpack and Replugged module
* Replaced hardcoded strings with i18n translations

[closes #331]


src/renderer/coremods/settings/pages/Addons.tsx
```diff
@@ -169,7 +169,7 @@ function Authors({ addon }: { addon: RepluggedPlugin | RepluggedTheme }) {
       ) : null}
       {author.github ? (
         <Tooltip
-          text="Open GitHub Profile"
+          text={Messages.REPLUGGED_ADDON_PROFILE_OPEN.format({ type: "GitHub" })}
```
asportnoy, line 172: Should there be an i18n key for GitHub

asportnoy (changes requested):

i18n/index.js
```diff
@@ -1,7 +1,39 @@
-require("fs")
-  .readdirSync(__dirname)
-  .filter((file) => file.endsWith(".json"))
-  .forEach((filename) => {
-    const moduleName = filename.split(".")[0];
-    exports[moduleName] = require(`${__dirname}/${filename}`);
-  });
+export default {
```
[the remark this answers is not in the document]
GriefMoDz: Seeing as I'm importing this file under the `renderer` context where the `fs` module isn't available, https://github.com/replugged-org/replugged/blob/53fba7e2e2c9d6e83e361cc875361d8f842271df/src/renderer/modules/i18n.ts#L4 I had to change this to hardcoded exports otherwise the project won't build.

The only other alternative that I see would be to fetch the i18n files via an IPC handler, I'm not sure if that's considered a better approach.

src/renderer/coremods/settings/lib.tsx
```diff
@@ -13,12 +17,12 @@ export const Section = ({
...
 }: {
   name?: string;
-  label: string;
+  label: string | LabelCallback;
   color?: string;
   elem: React.FunctionComponent;
   pos?: number;
 }): SectionType => ({
-  section: name || `REPLUGGED_${label.toUpperCase()}`,
+  section: name || `REPLUGGED_${(typeof label === "function" ? label() : label).toUpperCase()}`,
```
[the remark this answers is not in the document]
GriefMoDz: This is still required as it needs to resolve the label if it uses a callback (e.g. `() => Messages.REPLUGGED_GENERAL`).

src/renderer/coremods/settings/pages/Addons.tsx
GriefMoDz: I honestly believe that "GitHub" should remain hardcoded. Discord also doesn't have an i18n key for it, so I'd say this is reasonable.

src/main/ipc/i18n.ts
```diff
@@ -0,0 +1,8 @@
+import { ipcMain } from "electron";
+import { RepluggedIpcChannels } from "../../types";
+
+ipcMain.handle(RepluggedIpcChannels.GET_I18N_STRINGS, () => require("../../../i18n").default);
```
asportnoy, line 4: Could you import the folder above and then use that here and below?

asportnoy (commented):
One improvement we could make is to add a default value for any strings that are not yet translated. So if `REPLUGGED_CANCEL` was not available in some language, it could show a something (maybe the English translation or just the i18n key) instead of there being no text.

Overall this looks pretty good and I'll test it later.

GriefMoDz:
> One improvement we could make is to add a default value for any strings that are not yet translated. So if `REPLUGGED_CANCEL` was not available in some language, it could show a something (maybe the English translation or just the i18n key) instead of there being no text.

This has already been implemented, whereas it'll fallback to "English, US" for any translation string that isn't found.

asportnoy (approved):
Works as expected!

PR #356, opened 2023-01-24
Installer links

### LINKED ISSUE #256: Installer Link Coremod

asportnoy:
Example installer link: https://replugged.dev/install?identifier=asportnoy/mention-cache-fix&source=github

- Update to addon templates: the manifest is now included in the releases with the same name as the asar asset.
- Add `anyAddon` zod parser
- Improve updater to use the version in the manifests to tell if the version changed
- Set up installer flow (uses same code as updater)
- Clicking install link in Discord will trigger the installer flow
- Set up RPC to listen for events from installer links in the browser (code is updated to work with the rewrite)
- Add `Loader` component
- Fix types on `toast` module
- Fix unisntalling asar addons

Closes #256
Thanks to @Deepfried-Chips for getting this started!


src/renderer/coremods/installer/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+    return;
+  }
+
+  injector.before(exports, key, ([args]) => {
+    const { href } = args;
+    if (!href) return undefined;
+    const installLink = parseInstallLink(href);
+    if (!installLink) return undefined;
+
+    delete args.href;
```
12944qwerty, line 73: By deleting this, when you right click the link, it doesn't offer the same link context menu as it does with any other link. (this is the only suspect I can think of at least)

src/renderer/coremods/installer/util.ts
```diff
@@ -0,0 +1,244 @@
...
+      return themeManager.themes.has(data.manifest.id);
+  }
+}
+
+export async function loadNew(data: CheckResultSuccess): Promise<boolean> {
+  try {
+    switch (data.manifest.type) {
+      case "replugged-plugin":
+        await pluginManager.loadAll();
+        await pluginManager.start(data.manifest.id);
```
12944qwerty, line 71: enable already starts the plugin, don't do it twice?
```suggestion
```

12944qwerty (changes requested):

12944qwerty (approved):

PR #357, opened 2023-01-24
Quick CSS tweaks

asportnoy:
- Catch error if file doesn't exist
- Add button to open folder
- Create file on boot if it doesn't exist already


src/util.ts
```diff
@@ -32,3 +31,8 @@ export const CONFIG_PATHS = Object.fromEntries(
     return [name, path];
   }),
 ) as Record<typeof CONFIG_FOLDER_NAMES[number], string>;
+
+const QUICK_CSS_FILE = join(CONFIG_PATHS.quickcss, "main.css");
+if (!existsSync(QUICK_CSS_FILE)) {
+  writeFileSync(QUICK_CSS_FILE, "");
+}
```
12944qwerty, line 38: Shouldn't we create the files for all? I know this pr is for quick css but still :shrug:

src/main/ipc/quick-css.ts
```diff
@@ -6,7 +6,9 @@ import { CONFIG_PATHS } from "src/util";
 
 const CSS_PATH = join(CONFIG_PATHS.quickcss, "main.css");
 
-ipcMain.handle(RepluggedIpcChannels.GET_QUICK_CSS, () => readFile(CSS_PATH, { encoding: "utf-8" }));
+ipcMain.handle(RepluggedIpcChannels.GET_QUICK_CSS, () =>
+  readFile(CSS_PATH, { encoding: "utf-8" }).catch(() => ""),
```
12944qwerty, line 10: Shouldn't there be better errorhandling? Maybe make the file if it's not found?

src/util.ts
asportnoy: This is the only file we needed to make. No files to create for plugins and themes and we don't know the file names for settings (plus they'll be created when they're needed).

src/main/ipc/quick-css.ts
asportnoy: Already added that.

12944qwerty: That's on discord load... not here?

asportnoy: Discord load happens before this. Adding it here will make no difference.

12944qwerty (approved):

PR #359, opened 2023-01-25
Fix FormText component

E-boi:
get the correct module for the FormText component


asportnoy:
This doesn't seem to change anything on either Canary or Stable.

E-boi:
hm weird when I made this pr it would get some other component which would cause the crash but now it doesn't

asportnoy:
Gonna close this for now, we can revisit if the issue comes back.

E-boi:
The issue is happening again

asportnoy:
Still can't reproduce. Can you show the console output with and without this change? Also, what's your platform info?

E-boi:
wdym console output? the error or something else?

E-boi:
My platform info:
image

The error without this change:
image

Simplist code that i've been able to reproduce with (used for a setting page):
```js
import { components } from "replugged";
const { FormText } = components;

export default () => {
  return (
    <div>
      <FormText.DESCRIPTION>Hello world</FormText.DESCRIPTION>
    </div>
  );
};
```

asportnoy:
Can you log the component in the console though? Just `replugged.components.FormText`

E-boi:
The both are the same since FormText doesn't export the module itself

without the change:
image

with the change:
image

asportnoy (approved):
Still can't reproduce on stable or canary but I don't think this PR will hurt anything so I'll go ahead and approve

PR #360, opened 2023-01-25
Webpack by value

12944qwerty:
Get the module based on value string/regex.

### DIFF
--- a/src/renderer/modules/webpack/index.ts
+++ b/src/renderer/modules/webpack/index.ts
@@ -384,19 +384,21 @@ export namespace filters {
   /**
    * Get a module that has the given value on one of its exports
    * @param match The string to check the value against
-   * 
+   *
    * @remarks
    * Great for getting a module for a specific classname
    */
   export const byValue = (match: string | RegExp) => {
     return (m: RawModule) => {
       if (m.exports) {
-        return typeof match === "string" ? Object.values(m.exports).includes(match) : Object.values(m.exports).some(val => typeof val === "string" &&  match.test(val));
+        return typeof match === "string"
+          ? Object.values(m.exports).includes(match)
+          : Object.values(m.exports).some((val) => typeof val === "string" && match.test(val));
       } else {
-        return false
+        return false;
       }
-    }
-  }
+    };
+  };
 }
 
 // Async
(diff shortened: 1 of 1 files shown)


src/renderer/modules/webpack/index.ts
```diff
@@ -380,6 +380,25 @@ export namespace filters {
       return typeof match === "string" ? source.includes(match) : match.test(source);
     };
   };
+
+  /**
+   * Get a module that has the given value on one of its exports
+   * @param match The string to check the value against
+   *
+   * @remarks
+   * Great for getting a module for a specific classname
```
asportnoy, line 389: Can you put an example here?

src/renderer/modules/webpack/index.ts
```diff
@@ -380,6 +380,25 @@ export namespace filters {
...
+   */
+  export const byValue = (match: string | RegExp) => {
+    return (m: RawModule) => {
+      if (m.exports) {
+        return typeof match === "string"
+          ? Object.values(m.exports).includes(match)
+          : Object.values(m.exports).some((val) => typeof val === "string" && match.test(val));
+      } else {
+        return false;
+      }
```
asportnoy, line 399: Better to handle the edge cases first
```suggestion
      if (!m.exports) {
        return false
      }

      return typeof match === "string"
        ? Object.values(m.exports).includes(match)
        : Object.values(m.exports).some((val) => typeof val === "string" && match.test(val));
```

asportnoy (changes requested):

src/renderer/modules/webpack/index.ts
```diff
@@ -380,6 +380,29 @@ export namespace filters {
...
+   *
+   * @remarks
+   * Great for getting a module for a specific classname
+   */
+  export const byValue = (match: string | RegExp) => {
+    return (m: RawModule) => {
+      if (!m.exports) {
+        return false;
+      }
+      return typeof match === "string"
```
asportnoy, line 401: module exports is not guaranteed to be an object, so you should check for that here

asportnoy (changes requested):

src/renderer/modules/webpack/index.ts
```diff
@@ -395,7 +395,7 @@ export namespace filters {
    */
   export const byValue = (match: string | RegExp) => {
     return (m: RawModule) => {
-      if (!m.exports) {
+      if (!m.exports && typeof m.exports !== "object") {
```
asportnoy, line 398:
```suggestion
      if (!m.exports || typeof m.exports !== "object") {
```

asportnoy (changes requested):

asportnoy (approved):

PR #362, opened 2023-01-28
Fix exports being removed when loading missing plugins

asportnoy:


src/renderer/managers/plugins.ts
```diff
@@ -41,9 +41,10 @@ export function getExports(id: string): PluginExports | undefined {
 }
 
 function register(plugin: RepluggedPlugin): void {
+  const existingExports = plugins.get(plugin.manifest.id)?.exports;
```
12944qwerty, line 44: again, `getExports` does exist lolll

12944qwerty (approved):

src/renderer/managers/plugins.ts
asportnoy: `getExports` will throw if the plugin doesn't exist. We don't want that.

PR #364, opened 2023-01-28
addon build scripts

12944qwerty:
Addon build scripts.

In replugged dir, run `npm link`
In plugin/theme run `pnpm uninstall replugged && pnpm install path/to/replugged --save-dev && pnpm link replugged`

~~You can change the scripts or just run `build_replugged_plugin` directly from cmd. Since bundling is the same with theme and plugin, they were combined.~~

Changed to use subcommands. Added to package commands so that works too


bin/addonbundle.js
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env node
+
+const { join } = require("path");
```
asportnoy, line 3: Can you use the new version from https://github.com/replugged-org/plugin-template/commit/8d08a035d894ff4cf1b37b59673e3f23dae652fd#diff-56de87ca2ae95ee7c7a031bf487c87288b5dd206b66fed3ab6f551bf58081645

bin/pluginbuild.js
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env node
+/* eslint-disable consistent-return */
+
+const directory = process.cwd();
+const esbuild = require("esbuild");
+const { join } = require("path");
+const { cpSync, existsSync, mkdirSync, rmSync, writeFileSync } = require("fs");
+const _manifest = require(join(directory, "./manifest.json"));
+// const { PluginManifest } = require(join(process.cwd(), "node_modules/replugged/dist/types/addon"));
```
asportnoy, line 9:
```suggestion
```

bin/themebuild.js
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env node
+
+const dir = process.cwd();
+
+const { join } = require("path");
+const { Parcel } = require(join(dir, "node_modules/@parcel/core"));
```
asportnoy, line 6: Add this to the dev dependencies and import it properly

bin/addonbundle.js
asportnoy, line 3: We should use `import` instead of `require`.

asportnoy (changes requested):
This should be made into a proper subcommand. I'd imagine it would look like:
- `npx replugged build [--plugin|--theme]`
- `npx replugged bundle`

asportnoy:
Also, please make a PR to the plugin and theme templates with the required changes

bin/addonbundle.js
12944qwerty: It was using the new version lol

12944qwerty:
Subcommands now work like `replugged build <plugin|theme> [--watch] [--no-install]` and just `replugged bundle`

bin/replugged.js
```diff
@@ -0,0 +1,270 @@
+#!/usr/bin/env node
+/* eslint-disable consistent-return */
```
asportnoy, line 2: why

asportnoy (changes requested):
Add the `--production` option to minify and disable source maps. Should be used automatically when bundling (bundle should build beforehand and we should remove the build-and-bundle script in the templates)

bin/replugged.js
```diff
@@ -0,0 +1,270 @@
...
+
+function bundleAddon() {
+  const manifest = JSON.parse(readFileSync("dist/manifest.json", "utf-8"));
+  const outputName = `bundle/${manifest.id}.asar`;
+
+  if (!existsSync("bundle")) {
+    mkdirSync("bundle");
+  }
+  asar.createPackage("dist", outputName);
+  copyFileSync("dist/manifest.json", `${outputName}.json`);
```
asportnoy, line 19: This still needs to be updated to match the changes for the installer

asportnoy (changes requested):

bin/replugged.js
12944qwerty: Lines `34` and `46` give an eslint error because "no returning in arrow function". I'll change it to regular function ig

bin/replugged.js
```diff
@@ -10,26 +9,26 @@ const { join } = require("path");
...
+function buildPlugin({ watch, noInstall, production }) {
   const manifest = require(join(directory, "./manifest.json"));
   const CHROME_VERSION = "91";
   const REPLUGGED_FOLDER_NAME = "replugged";
   const globalModules = {
     name: "globalModules",
     setup: (build) => {
       build.onResolve({ filter: /^replugged.+$/ }, (args) => {
-        if (args.kind !== "import-statement") return;
+        if (args.kind !== "import-statement") return {};
```
asportnoy, line 31:
```suggestion
        if (args.kind !== "import-statement") return undefined;
```

bin/replugged.js
```diff
@@ -41,7 +40,7 @@ function buildPlugin({ watch, noInstall }) {
       });
   
       build.onResolve({ filter: /^replugged$/ }, (args) => {
-        if (args.kind !== "import-statement") return;
+        if (args.kind !== "import-statement") return {};
```
asportnoy, line 43:
```suggestion
        if (args.kind !== "import-statement") return undefiend;
```

asportnoy (commented):
Add `./bin` to the lint commands in package and format this file.

bin/replugged.js
```diff
@@ -0,0 +1,297 @@
...
+  const main = manifest.main || "src/main.css";
+  const splash = manifest.splash || (existsSync("src/splash.css") ? "src/splash.css" : undefined);
+
+  const mainBundler = new Parcel({
+    entries: main,
+    defaultConfig: "@parcel/config-default",
+    targets: {
+      main: {
+        distDir: "dist",
+        distEntry: "main.css",
```
asportnoy, line 164: For production, set `sourceMap` to false and `optimize` to true

bin/replugged.js
```diff
@@ -0,0 +1,297 @@
...
+  copyFileSync,
+  existsSync,
+  mkdirSync,
+  readFileSync,
+} = require("fs");
+const esbuild = require("esbuild");
+const { join } = require("path");
+
+function bundleAddon() {
+  const manifest = JSON.parse(readFileSync("dist/manifest.json", "utf-8"));
```
asportnoy, line 19: Delete the dist folder here so that maps that are already built won't be included in the bundle

bin/replugged.js
```diff
@@ -0,0 +1,297 @@
+#!/usr/bin/env node
```
asportnoy, line 1: Shebang is not needed and causes issues in the editor so let's remove it.

asportnoy (changes requested):

asportnoy (approved):
image

PR #366, opened 2023-01-30
Slider component

cafeed28:
Example:
```tsx
<Slider
  markers={[16, 24, 32, 40, 48, 56, 64, 80, 96]}
  stickToMarkers={true}
  defaultValue={48}
  initialValue={32}
  minValue={16}
  maxValue={96}
  onMarkerRender={(e: number) => {
    return `${e}px`;
  }}
/>
```
image

### DIFF
--- a/src/renderer/modules/components/Slider.tsx
+++ b/src/renderer/modules/components/Slider.tsx
@@ -1,6 +1,7 @@
+import { ReactComponent } from "src/types";
 import { filters, waitForModule } from "../webpack";
 
-interface SliderProps {
+interface SliderCompProps {
   disabled?: boolean;
   markers?: number[];
   stickToMarkers?: boolean;
@@ -10,16 +11,30 @@ interface SliderProps {
   maxValue?: number;
   mini?: boolean;
   style?: React.CSSProperties;
+  className?: string;
   onValueChange?: (e: number) => void;
   asValueChanges?: (e: number) => void;
   onValueRender?: (e: number) => string;
   onMarkerRender?: (e: number) => string;
 }
 
-export type SliderType = React.ComponentType<SliderProps> & {
-  defaultProps: SliderProps;
+type SliderCompType = React.ComponentType<SliderCompProps> & {
+  defaultProps: SliderCompProps;
 };
 
-export default (await waitForModule(filters.bySource(".moveGrabber=")).then((mod) =>
+const SliderComp = (await waitForModule(filters.bySource(".moveGrabber=")).then((mod) =>
   Object.values(mod).find((x) => "defaultProps" in x && "stickToMarkers" in x.defaultProps),
-)) as SliderType;
+)) as SliderCompType;
+
+interface SliderProps extends SliderCompProps {
+  value?: number;
+  onChange?: (e: number) => void;
+}
+
+type SliderType = ReactComponent<SliderProps>;
+
+export default ((props) => {
+  return (
+    <SliderComp {...props} initialValue={props.value} onValueChange={props.onChange}></SliderComp>
+  );
+}) as SliderType;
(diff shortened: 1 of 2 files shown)


src/renderer/modules/components/Slider.tsx
```diff
@@ -0,0 +1,25 @@
...
+  initialValue?: number;
+  defaultValue?: number;
+  minValue?: number;
+  maxValue?: number;
+  mini?: boolean;
+  style?: React.CSSProperties;
+  onValueChange?: (e: number) => void;
+  asValueChanges?: (e: number) => void;
+  onValueRender?: (e: number) => string;
+  onMarkerRender?: (e: number) => string;
```
asportnoy, line 16: Missing the `className` property

src/renderer/modules/components/Slider.tsx
```diff
@@ -0,0 +1,25 @@
+import { filters, waitForModule } from "../webpack";
+
+interface SliderProps {
+  disabled?: boolean;
+  markers?: number[];
+  stickToMarkers?: boolean;
+  initialValue?: number;
+  defaultValue?: number;
```
asportnoy, line 8: `defaultValue` doesn't exist

asportnoy (changes requested):
Can you add aliases for the props `initialValue` as `value` and `onValueChange` as `onChange`? (Create a separate component that wraps the one from Discord and apply the values on the original object). That way, it'll work better with the `useSetting` hook.

src/renderer/modules/components/Slider.tsx
cafeed28: `defaultValue` is green marker

image

src/renderer/modules/components/index.ts
```diff
@@ -91,6 +91,11 @@ export type { TextInputType };
 export let TextInput: TextInputType;
 importTimeout("Input", import("./TextInput"), (mod) => (TextInput = mod.default));
 
+import type { SliderType } from "./Slider";
+export type { SliderType };
+export let Slider: SliderType;
+importTimeout("Input", import("./Slider"), (mod) => (Slider = mod.default));
```
FedeIlLeone, line 97: Shouldn't "Input" be changed to "Slider"?

src/renderer/modules/components/Slider.tsx
```diff
@@ -0,0 +1,40 @@
...
+interface SliderProps extends SliderCompProps {
+  value?: number;
+  onChange?: (e: number) => void;
+}
+
+export type SliderType = ReactComponent<SliderProps>;
+
+export default ((props) => {
+  return (
+    <SliderComp {...props} initialValue={props.value} onValueChange={props.onChange}></SliderComp>
```
asportnoy, line 38: Move the spread to the end so people can still use the original names

asportnoy (changes requested):

src/renderer/modules/components/Slider.tsx
cafeed28: Done

src/renderer/modules/components/index.ts
cafeed28: Fixed

src/renderer/modules/components/Slider.tsx
cafeed28: Fixed in [this commit](https://github.com/replugged-org/replugged/pull/366/commits/42ff71b350430157d78a50f5bce59b144e07d1cb#diff-ba27653ecb680e72e999facdb29940cc49686074aee6280329274baf5d104122R14)

asportnoy (commented):
Code looks good, will test later today

asportnoy (approved):

Issue #368, opened 2023-02-01
Components 


Tharki-God:
### Describe the feature request

i want some components added to replugged.components because they are really common, for example a slider or dropdown 
as for the code: [Slider](https://github.com/Tharki-God/ShowNames/blob/main/src/Components/Slider.jsx),
 [SettingsGroup](https://github.com/Tharki-God/ShowHiddenChannels/tree/main/src/Components) a group of setting that can minimised or collapsed to just a string of text,
and [dropdowns](https://github.com/Tharki-God/PremiumScreenShare/blob/main/src/Components/Dropdown.jsx) and ig this is already in pull request but still [radiogroup](https://github.com/Tharki-God/ShowHiddenChannels/blob/main/src/Components/RadioGroup.jsx)

### Why should this be added?

because these are commonly used components

### Additional information

please unban me from the server pls

### Please confirm the following

- [X] This is a feature request for Replugged itself, not a feature request for a plugin or theme
- [X] [I searched the issues](https://github.com/replugged-org/replugged/issues) to make sure this feature has not been requested before.

asportnoy:
Slider is being added in #366
SettingsGroup is already implemented (`replugged.components.Category`)
Radio is being added in #354

Tharki-God:
That leaves dropdowns

PR #369, opened 2023-02-02
Select component

FedeIlLeone:
Adds a `Select` component.

Example:
```tsx
<Select
  options={[
    { label: "Option 1", value: "opt1" },
    { label: "Option 2", value: "opt2" },
    { label: "Option 3", value: "opt3" },
  ]}
  onSelect={(e) => setSelected(e)}
  isSelected={(e) => e === selected}
  onClear={() => setSelected("")}
  clearable
  placeholder="Placeholder text"
/>
```

image

Closes: #368

### DIFF
--- a/src/renderer/modules/components/RadioItem.tsx
+++ b/src/renderer/modules/components/RadioItem.tsx
@@ -4,7 +4,7 @@ import { Divider, FormItem, FormText, Text } from ".";
 
 interface RadioOptionType {
   name: string;
-  value: string | number;
+  value: string;
   desc?: string;
   disabled?: boolean;
   color?: string;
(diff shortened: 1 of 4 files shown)


asportnoy (changes requested):
Let's add a wrapper similar to this so we can use more traditional `value` / `onChange` (and work with useSetting too)
Screenshot on 2023-02-02 at 17 57 41

FedeIlLeone:
Ok done, I think thats what you meant.

asportnoy (changes requested):
2 requests: Can we make the value type string only instead of string or number? And can you add a `SelectItem` wrapper like with radio? Looks good otherwise!

FedeIlLeone:
Done
image

asportnoy (approved):
Thanks!

PR #371, opened 2023-02-04
Fix FormItem, ButtonItem and Category

FedeIlLeone:
- Fixes `FormItem` component, with new props as well.
- Fixes `ButtonItem`, with the use of Flex. It was very messed up with centering the button.
- Fixes `Category`, added children type and doesn't break anymore if title/note is long.


asportnoy (changes requested):
Per discussion in Discord, we should use the prop for description instead of `<FormText.DESCRIPTION>`

asportnoy (approved):

PR #372, opened 2023-02-05
Wait for props

### LINKED ISSUE #314: Proposal: waitForProps
### Describe the feature request

Add a method to the webpack module to simplify the process of lazily fetching a module by props.

### Why should this be added?

Using waitForModule, filters.byProps, and getExportsForProps is cumbersome, but it seems to have become common practice in current plugins and the mod itself due to the lack of a better way. A method to handle all of that in one function call would make lazy webpack searches more accessible.

### Additional information

An example from renderer/modules/webpack/common/lodash.ts:

```ts
import { filters, getExportsForProps, waitForModule } from "..";

import type Lodash from "lodash";

const lodash = getExportsForProps(await waitForModule(filters.byProps("debounce")), [
  "debounce",
]) as unknown as typeof Lodash;
```

This could be simplified to:

```ts
import { waitByProps } from "..";

import type Lodash from "lodash";

const lodash = await waitByProps("debounce") as unknown as typeof Lodash;
```

- [X] This is a feature request for Replugged itself, not a feature request for a plugin or theme
- [X] [I searched the issues](https://github.com/replugged-org/replugged/issues) to make sure this feature has not been requested before.

asportnoy:
Add `waitForProps` webpack helper. Also removed JSDoc comments for each overload because they were just there for typedoc and they're not necessary anymore since we're switching to our own guide.

Closes #314 

Relevant changes start at line 567, you can ignore anything above that.

### DIFF
--- a/src/renderer/modules/webpack/index.ts
+++ b/src/renderer/modules/webpack/index.ts
@@ -567,19 +567,27 @@ export function getByProps<
 export function waitForProps<
   P extends string = string,
   T extends ModuleExportsWithProps<P> = ModuleExportsWithProps<P>,
->(props: P[], options: { raw?: false }): Promise<T | undefined>;
+>(props: P[], options: { raw?: false; timeout?: undefined }): Promise<T>;
+export function waitForProps<
+  P extends string = string,
+  T extends ModuleExportsWithProps<P> = ModuleExportsWithProps<P>,
+>(props: P[], options: { raw?: false; timeout: number }): Promise<T | undefined>;
 export function waitForProps<P extends string = string, T extends RawModule = RawModule>(
   props: P[],
-  options: { raw?: true },
+  options: { raw?: true; timeout?: undefined },
+): Promise<T>;
+export function waitForProps<P extends string = string, T extends RawModule = RawModule>(
+  props: P[],
+  options: { raw?: true; timeout: number },
 ): Promise<T | undefined>;
 export function waitForProps<
   P extends string = string,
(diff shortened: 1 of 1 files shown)


src/renderer/modules/webpack/index.ts
```diff
@@ -737,6 +564,48 @@ export function getByProps<
...
+export function waitForProps<
+  P extends string = string,
+  T extends ModuleExportsWithProps<P> | RawModule = ModuleExportsWithProps<P> | RawModule,
+>(props: P[], options?: WaitForOptions): Promise<T>;
+export function waitForProps<
+  P extends string = string,
+  T extends ModuleExportsWithProps<P> = ModuleExportsWithProps<P>,
+>(...props: P[]): T;
+
+/**
```
12944qwerty, line 584: deletes all tsdoc... adds two docs anyway

12944qwerty (approved):
Can we just make the waitFor for source too?

src/renderer/modules/webpack/index.ts
asportnoy: It doesn't delete them _all_, it just moves them to the actual function instead of the overloads. This is intentional.

asportnoy:
> Can we just make the waitFor for source too?

There isn't any special handling needed for it; `waitForModule(filters.bySource(xxx))` does the exact same thing as `waitForSource` would. `waitForProps` does some additional processing on the module returned (`getExportForProps`).

asportnoy:
Thought I merged this 🤦

Issue #376, opened 2023-02-09
SliderItem


Tharki-God:
### Describe the feature request

yeah there is slider added but its as raw component, please add a component with form title and note.

### Why should this be added?

to have consistency with other components

### Please confirm the following

- [X] This is a feature request for Replugged itself, not a feature request for a plugin or theme
- [X] [I searched the issues](https://github.com/replugged-org/replugged/issues) to make sure this feature has not been requested before.

Tharki-God:
also another request is that make the title as a prop on the component instead of children, it would make writing code cleaner tbh, not a big issue to make another issue out of so just leaving this as component here

asportnoy:
You can wrap it in a FormItem yoursel

Tharki-God:
i am already doing that , just wanted it to get added

Tharki-God:
i dont know ts, so cant add types thats why didnt made pr and too lazy to do it anyways sorry.

PR #382, opened 2023-02-12
New components and types

FedeIlLeone:
This PR is mainly done to improve types (for docs/guide) and add some more components.

- Add `Checkbox` and `CheckboxItem` component. Example:
```tsx
<Checkbox
  value={sw}
  onChange={() => setSw(!sw)}
  type={Checkbox.Types.INVERTED}
/>
<CheckboxItem value={sw} onChange={() => setSw(!sw)} type={Checkbox.Types.INVERTED}>
  Title placeholder
</CheckboxItem>
```
image
- Add `TextArea` component. Example:
```tsx
<TextArea placeholder="Placeholder text" />
```
image
- Add `SliderItem` component (#376 ). Example:
```tsx
<SliderItem
  markers={[16, 24, 32, 40, 48, 56, 64, 80, 96]}
  stickToMarkers={true}
  defaultValue={48}
  initialValue={32}
  minValue={16}
  maxValue={96}
  onMarkerRender={(e: number) => `${e}px`}
  note="Note placeholder.">
  Title Placeholder
</SliderItem>
```
image
- Add missing types on `Flex`, `FormText`, `Modal`, `Slider`, `SwitchItem` and `TextInput` components.
- Add `noteStyle` type on `FormItem` component.
- Fix `style` on `RadioItem`, `SelectItem` and `Slider` component.
- Fix `note` on `ButtonItem` and `Category` component.
- Fix import names.

Closes: #376


src/renderer/modules/components/CheckboxItem.tsx
```diff
@@ -0,0 +1,46 @@
...
+};
+
+export type CheckboxItemType = React.FC<React.PropsWithChildren<CheckboxProps>>;
+
+export const Checkbox = await waitForModule(filters.bySource(".getInputMode")).then((mod) =>
+  Object.values(mod).find((x) => "defaultProps" in x && "displayOnly" in x.defaultProps),
+);
+
+export const CheckboxItem = (props: React.PropsWithChildren<CheckboxProps>): React.ReactElement => {
+  return (
```
asportnoy, line 37: Should this have a note as well?

src/renderer/modules/components/FormItem.tsx
```diff
@@ -26,23 +26,26 @@ const classes = await waitForModule<Record<"dividerDefault", string>>(filters.by
 interface FormItemProps extends FormItemCompProps {
   note?: string;
   notePosition?: "before" | "after";
+  noteStyle?: React.CSSProperties;
```
asportnoy, line 29: Could you add `noteClassName` too?

src/renderer/modules/components/RadioItem.tsx
```diff
@@ -39,6 +38,7 @@ export const Radio = (await waitForModule(filters.bySource(radioStr)).then((mod)
 
 interface RadioItemProps extends RadioProps {
   note?: string;
+  style?: React.CSSProperties;
```
asportnoy, line 41: I think style can be used on the original component. Did you test that? (applies to other files as well)

src/renderer/modules/components/RadioItem.tsx
```diff
@@ -47,11 +47,10 @@ export const RadioItem = (props: React.PropsWithChildren<RadioItemProps>): React
   return (
     <FormItem
       title={props.children}
-      style={{ marginBottom: 20 }}
+      style={{ marginBottom: 20, ...props.style }}
```
asportnoy, line 50: We should use a different property for this so we can style it separately from the component itself (applies to other files as well)

src/renderer/modules/components/TextArea.tsx
```diff
@@ -0,0 +1,33 @@
+import { filters, waitForModule } from "../webpack";
```
asportnoy, line 1: Isn't this the same as TextInput?

asportnoy (changes requested):
Few minor things

src/renderer/modules/components/CheckboxItem.tsx
FedeIlLeone: I don't know, I couldn't find any examples made by them. There is only in the automod section a checkbox with a note, but it's totally different from how it is used in any other place.

src/renderer/modules/components/FormItem.tsx
FedeIlLeone: Done

src/renderer/modules/components/RadioItem.tsx
FedeIlLeone: Yeah I did test. Where I removed the style prop and moved it to the FormItem then the original component doesn't have a style prop.

src/renderer/modules/components/TextArea.tsx
FedeIlLeone: Not really. TextInput should be made to accept small values, TextArea can be longer (like descriptions) and of course a different customization props than TextInput. I don't know if it would be useful to keep this component built-in.

src/renderer/modules/components/RadioItem.tsx
asportnoy: So to confirm, setting the style doesn't change anything?

src/renderer/modules/components/TextArea.tsx
asportnoy: I mean that I think it's the same component as input (same selector and everything). Is there just an option to make it a textarea vs input? Or am I missing something.

src/renderer/modules/components/RadioItem.tsx
FedeIlLeone: exactly

src/renderer/modules/components/TextArea.tsx
FedeIlLeone: Nope, they are two different components, nothing in common between them.

asportnoy (approved):

Issue #384, opened 2023-02-13
Move webpack common into modules folder


asportnoy:
The components are also in that directory, so I think it makes sense to put common there as well.
Since there will only be one file left in the folder, you can move that into the modules folder as well and rename it to `webpack.ts`.

Of course, make sure to update the paths for any references to those files or folders, including in `tsconfig.json` and make sure this doesn't break any of Replugged's features or plugins.

Kirito-Excalibur:
Can you assign me this?

lexisother:
@Kirito-Excalibur Done.

PR #388, opened 2023-02-14
Moved webpack common into modules folder

Kirito-Excalibur:
Solves issue #384 

Moved the required folders and also refactored all imports.


src/renderer/coremods/badges/badge.tsx
```diff
@@ -1,6 +1,6 @@
-import { filters, getByProps, waitForModule } from "../../modules/webpack";
-import { React } from "@common";
```
asportnoy, line 2: Why did you get rid off these? We would prefer `@common` over the full path

src/renderer/modules/common/api.ts
```diff
@@ -1,4 +1,4 @@
-import { filters, getExportsForProps, waitForModule } from "..";
+import { filters, getExportsForProps, waitForModule } from "./webpack";
```
asportnoy, line 1: Are you sure this is correct? I think it should be `../webpack`

asportnoy (changes requested):

src/renderer/modules/common/api.ts
Kirito-Excalibur: I found this to be correct.

asportnoy:
@Kirito-Excalibur did you accidentally close this?

Kirito-Excalibur:
Yeah😅😅😅

PR #390, opened 2023-02-15
Moved webpack common into modules folder

Kirito-Excalibur:
Solves issue #384.
Moved common and the otherfile from webpack to modules folder and also renamed the file `webpack.ts`
Refactored as required.


asportnoy:
Code looks good, I'll test locally later and then approve if there are no issues

asportnoy (approved):

PR #399, opened 2023-02-20
Components tweaks

FedeIlLeone:
Made this PR for https://github.com/replugged-org/guide/pull/12 .

- Redo `Button` and `ButtonItem` with new types (a deprecated component was being used before).
- Redo `Tooltip` with new types and enums.
- Show visually if `Category` is disabled. Same thing to other Item components.
- Rename `Notice`'s `NoticeTypes` enum to `Types`.
- Types changes.
- Use `submitting` prop instead of `Loader` in Updater page.
- Fix `SliderItem` if used with `Category`. The title disappeared if the category was opened a second time.


src/renderer/coremods/settings/pages/Updater.tsx
```diff
@@ -132,13 +132,10 @@ export const Updater = (): React.ReactElement => {
...
             className="replugged-updater-check"
             onClick={checkForUpdates}
-            disabled={checking || isAnyUpdating || isAnyComplete}
-            color={checking ? Button.Colors.PRIMARY : Button.Colors.BRAND}>
-            {checking ? (
-              <Loader type={Loader.Type.PULSING_ELLIPSIS} />
-            ) : (
-              Messages.REPLUGGED_UPDATES_CHECK
-            )}
+            disabled={isAnyComplete}
```
asportnoy, line 135:
```suggestion
            disabled={isAnyUpdating || isAnyComplete}
```

src/renderer/coremods/settings/pages/Updater.tsx
```diff
@@ -147,13 +144,10 @@ export const Updater = (): React.ReactElement => {
...
-            disabled={isAnyUpdating || isAllComplete}
-            color={isAllUpdating ? Button.Colors.PRIMARY : Button.Colors.BRAND}>
-            {isAllUpdating ? (
-              <Loader type={Loader.Type.PULSING_ELLIPSIS} />
-            ) : (
-              Messages.REPLUGGED_UPDATES_UPDATE_ALL
-            )}
+            disabled={isAllComplete}
+            color={isAllUpdating ? Button.Colors.PRIMARY : Button.Colors.BRAND}
+            submitting={isAllUpdating}>
```
asportnoy, line 149:
```suggestion
            submitting={isAnyUpdating}>
```

src/renderer/modules/components/Loader.tsx
```diff
@@ -21,7 +21,6 @@ export type LoaderType = React.ComponentType<LoaderProps> & {
   Type: typeof TYPES;
 };
 
-// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
-export default (await waitForModule(filters.bySource('"wanderingCubes"')).then((mod) =>
+export default await waitForModule(filters.bySource('"wanderingCubes"')).then((mod) =>
   Object.values(mod).find((x) => typeof x === "function"),
-)) as LoaderType;
+);
```
asportnoy, line 26: Type here needs to be put back. Think this is getting caused by some ESLint version thing but idk.

src/renderer/modules/components/Notice.tsx
```diff
@@ -9,20 +9,19 @@ const NoticeTypes = {
...
-  messageType: (typeof NoticeTypes)[keyof typeof NoticeTypes];
+  messageType: (typeof Types)[keyof typeof Types];
   textColor?: string;
   textVariant?: string;
   className?: string;
 }
 
 export type NoticeType = React.ComponentType<NoticeProps> & {
-  NoticeTypes: typeof NoticeTypes;
+  Types: typeof Types;
```
asportnoy, line 19: This would normally be a breaking change but I think it's ok here since we haven't announced or released the change yet

src/renderer/modules/components/Notice.tsx
```diff
@@ -9,20 +9,19 @@ const NoticeTypes = {
...
 };
 
-// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
-const Notice = (await waitForModule(filters.bySource(/.\.messageType/)).then((mod) =>
+const Notice = await waitForModule(filters.bySource(/.\.messageType/)).then((mod) =>
   Object.values(mod).find((x) => typeof x === "function"),
-)) as NoticeType;
-Notice.NoticeTypes = NoticeTypes;
+);
```
asportnoy, line 24: Same here, type needs to be put back.

src/renderer/modules/components/Tooltip.tsx
```diff
@@ -1,15 +1,42 @@
+import type { ObjectExports } from "../../../types";
 import type { HTMLAttributes } from "react";
 import { filters, getFunctionBySource, waitForModule } from "../webpack";
 
+const Aligns = {
+  BOTTOM: "bottom",
+  CENTER: "center",
```
asportnoy, line 7: Does `WINDOW_CENTER` need to be added here?

asportnoy (changes requested):

src/renderer/modules/components/Tooltip.tsx
FedeIlLeone: `WINDOW_CENTER` is for positions

src/renderer/modules/components/Notice.tsx
FedeIlLeone: `Notice` comp hasn't been released yet so I think it's ok.

src/renderer/modules/components/Loader.tsx
FedeIlLeone: honestly I don't see any problem without that, as the type is specified when exporting in index.ts

asportnoy: @FedeIlLeone If we import it directly inside the mod, it needs to have types

asportnoy (approved):

Issue #385, opened 2023-02-13
error in plugging discord


NotPhantomX:
### Describe the bug

In src/util.ts configPathFn returns /root/.config/replugged instead of /home/{username}/.config/replugged causing an error since the .config folder in /root/ does not exist. The cause seems to be process.env.XDG_CONFIG_HOME being set to /root at some point. Using bash instead of zsh or fish sets process.env.HOME to /root as well.

### Reproduction steps

1. sudo pnpm run plug --production canary

### Expected behavior

expected replugged to succeed in plugging discord

### Actual behavior

tries to create /root/.config/replugged and fails to create the path since /root/.config also does not exist.

### Additional information

```
sudo pnpm plug --production canary                         

> replugged@4.0.0-beta0.27 plug /home/phantomx/.replugged
> tsx scripts/inject/index.ts inject --no-exit-codes "--production" "canary"

node:fs:1394
  handleErrorFromBinding(ctx);
  ^

Error: ENOENT: no such file or directory, mkdir '/root/.config/replugged'
    at mkdirSync (node:fs:1394:3)
    at Object.<anonymous> (/home/phantomx/.replugged/src/util.ts:22:3)
    at Module._compile (node:internal/modules/cjs/loader:1120:14)
    at Object.v (/home/phantomx/.replugged/node_modules/.pnpm/@esbuild-kit+cjs-loader@2.3.3/node_modules/@esbuild-kit/cjs-loader/dist/index.js:1:933)
    at Module.load (node:internal/modules/cjs/loader:998:32)
    at Module._load (node:internal/modules/cjs/loader:839:12)
    at Module.require (node:internal/modules/cjs/loader:1022:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/home/phantomx/.replugged/scripts/inject/injector.ts:7:30)
    at Module._compile (node:internal/modules/cjs/loader:1120:14) {
  errno: -2,
  syscall: 'mkdir',
  code: 'ENOENT',
  path: '/root/.config/replugged'
}

Node.js v18.7.0
 ELIFECYCLE  Command failed with exit code 1.
```
Operating System: Fedora 37
Discord Version: Canary 173897 (8aa1dde)

### Please confirm the following

- [X] I checked if Discord and Replugged are up to date
- [X] I made sure this is not related to an external plugin or a theme
- [X] [I searched the issues](https://github.com/replugged-org/replugged/issues) to make sure this bug has not been reported before.

[...]

PR #408, opened 2023-03-02
Fix home path for sudo

Vap0r1ze:
This fixes the injector failing as sudo due to this part of src/util.ts
```ts
if (!existsSync(CONFIG_PATH)) {
  mkdirSync(CONFIG_PATH);
}
```
Which is ran before the the inject script can overwrite `process.env.HOME`, so it tries to create a folder in `/root/.config` which might not exist and is the wrong folder anyways

The fix checks for `SUDO_USER` or `DOAS_USER` which are always set by the sudo/doas util, and finds the home directory from `/etc/passwd` (or whetever ur passwd file is)

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -14,10 +14,18 @@ export const configPathFn = (): string => {
       if (process.env.XDG_CONFIG_HOME) {
         return join(process.env.XDG_CONFIG_HOME, REPLUGGED_FOLDER_NAME);
       }
-      if (realUser) {
-        // Get the home directory of the sudo user from /etc/passwd
-        const homeDir = execSync(`getent passwd ${realUser}`).toString('utf-8').split(':')[5];
-        return join(homeDir, ".config", REPLUGGED_FOLDER_NAME);
+      try {
+        if (realUser) {
+          // Get the home directory of the sudo user from /etc/passwd
+          const homeDir = execSync(`getent passwd ${realUser}`, {
+            stdio: [null, null, "ignore"],
+          })
+            .toString("utf-8")
+            .split(":")[5];
+          return join(homeDir, ".config", REPLUGGED_FOLDER_NAME);
+        }
+      } catch {
+        console.error("Could not get home directory of sudo/doas user. Falling back to $HOME.");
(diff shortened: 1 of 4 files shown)


src/util.ts
```diff
@@ -12,6 +14,11 @@ export const configPathFn = (): string => {
       if (process.env.XDG_CONFIG_HOME) {
         return join(process.env.XDG_CONFIG_HOME, REPLUGGED_FOLDER_NAME);
       }
+      if (realUser) {
+        // Get the home directory of the sudo user from /etc/passwd
+        const homeDir = execSync(`getent passwd ${realUser}`).toString('utf-8').split(':')[5];
```
asportnoy, line 19: Needs some error handling. Maybe throw if this either returns nothing or the result doesn't exist on the system?

asportnoy (changes requested):

asportnoy:
Also prettier needs to be run

src/util.ts
Vap0r1ze: If the result doesn't exist on the system then that's a problem with your system, but it could be empty so I should handle that

src/util.ts
```diff
@@ -12,6 +14,19 @@ export const configPathFn = (): string => {
...
         return join(process.env.XDG_CONFIG_HOME, REPLUGGED_FOLDER_NAME);
       }
+      try {
+        if (realUser) {
+          // Get the home directory of the sudo user from /etc/passwd
+          const homeDir = execSync(`getent passwd ${realUser}`, {
+            stdio: [null, null, "ignore"],
+          })
+            .toString("utf-8")
+            .split(":")[5];
```
asportnoy, line 24: Should still verify this has a banks value and it is a valid path. If parsing fails, we want it to stop.

Vap0r1ze: i highly doubt you could even run a unix system if this is misconfigured, but i handled it anyway

asportnoy:
Confirmed that this PR fixes #385.

asportnoy (approved):

PR #411, opened 2023-03-04
Remove i18n overrides

FedeIlLeone:
Removes i18n overrides, no longer needed as Discord now supports UK language.


asportnoy (changes requested):
String deletions should be done through Weblate to avoid git conflicts

FedeIlLeone:
👍🏻 I removed them from Weblate

asportnoy (approved):

PR #386, opened 2023-02-13
Reload prompt for plaintextPatches

bop34:
Closes #377

PR #290, opened 2022-12-22
Remove the use of multiple if's on Coremod badges

Wolfkid200444:

PR #412, opened 2023-03-04
style: conventional commits

asportnoy:
Set up the repository for conventional commits
- Add actions to check the PR title and commit messages (commit messages are only checked on main)
- Add husky commit hook to warn users if their messages are invalid
> Disabled by default, run `pnpm run cfg:lint-commit-msg:true` if you want this
- Recommend conventional commits vscode extension

Also added commit hook for general linting while I was at it


Vendicated:
Please do not add a pre commit hook like this. It's inconvenient and annoying and noone wants to make conventional commits inside a pull request where you will make dozens of commits that will get squashed anyway. You can already enforce a commit style by just properly squashing and renaming prs

lexisother:
Thank you for putting into words what I didn't want to.

asportnoy:
@Vendicated @lexisother per your feedback, I made the GitHub action for commits only trigger on main and I removed the hook. The hook is still there and can be enabled with `pnpm run cfg:lint-commit-msg:true` for users (like me) who want to use conventional commits.

asportnoy:
As for the general lint hook, there have been a bunch of cases where someone just forgets to lint, so this will catch those and remove unnecessary commits for fixing that (or a bunch of extra commits if they fail to). If you are intending to push while TypeScript is still failing for whatever reason, you can just use `--no-verify` to disable it for that one commit.

Vendicated:
> As for the general lint hook, there have been a bunch of cases where someone just forgets to lint

in which case github actions will fail and notify them of their failure to lint, so the hook is still pointless and annoying

asportnoy:
> > As for the general lint hook, there have been a bunch of cases where someone just forgets to lint
> 
> in which case github actions will fail and notify them of their failure to lint, so the hook is still pointless and annoying

Not always the case. See #410, #408, #290 for example. Also see #386 where there were 4 different commits to try and fix the lint.

I think it's much better to run the linter before they commit and I think the situation you're describing is very infrequent. It it becomes an issue then we can revisit but I doubt that will be the case.

colin273 (changes requested):
Contributing to Replugged should not be any more needlessly complicated than it already is. Design by committee is enough of a drag on progress; we do not need this extra wrinkle that works great for exactly one person with very limited free time out of the entire team actively working on the mod itself.

Blocking merge, pending closing this PR without merging at all.

asportnoy:
@colin273 the only additional requirement is the title, which we can easily edit ourselves if needed.

PR #422, opened 2023-03-09
feat: add notice and percentages in language settings

FedeIlLeone:
- Notice in language settings with a reference to the Weblate site.
- Percentages that indicate how much languages ​​are translated.
- `FormNotice` component.

image


lexisother (approved):
LGTM.

asportnoy (commented):
Screenshot on 2023-03-11 at 14 18 16
I think this looks better. Add `background-secondary` as the background color.

asportnoy (approved):
per internal discussion

PR #430, opened 2023-03-17
Adds shebang to replugged executable

Allvaa:
The executable failed to run on Linux when I tried to run a command to build plugin.
image

Adding a shebang line fixed the problem. Now I can run the command with no error.
image

### DIFF
--- a/bin/replugged.mjs
+++ b/bin/replugged.mjs
@@ -1,3 +1,5 @@
+#!/usr/bin/env node
+
 const directory = process.cwd();
 import asar from "@electron/asar";
 import { Parcel } from "@parcel/core";


asportnoy:
I think there was a reason we removed this. Let me see if there is another workaround.

asportnoy (changes requested):
Blocking merge for now

Allvaa:
> [a6c7599](https://github.com/replugged-org/replugged/commit/a6c75993653c46eca80a281e786078733daf8baa)

Anyway why was it not needed?
Also I think issues in the editor are relative, I don't see any issue in my editor with shebang

asportnoy (approved):
I think editor issues were the initial issue and it worked without it (still does for me at least). But it doesn't seem to be having issues now so I'll approve.

PR #435, opened 2023-03-25
Add dev companion coremod

lexisother:
Adds support for Ven's VSCode companion plugin.

Allows testing of module finds:
image

And testing of plaintext patches:
image


src/renderer/coremods/devCompanion/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+  }
+}
+
+function search(find: string | RegExp): string[] {
+  return Object.values(sourceStrings).filter((s) => {
+    if (typeof find === "string") {
+      return s.includes(find);
+    }
+    if (find instanceof RegExp) {
+      return find.test(s);
```
Vendicated, line 89: if the regex is global, this will have unexpected results because global regexes keep state for test and exec. either use something like `!!s.match(find)` or reset the regex:
```suggestion
      const matches = find.test(s);
      find.lastIndex = 0;
      return matches;
```

colin273 (approved):

PR #436, opened 2023-03-25
feat: addon watcher

### LINKED ISSUE #361: Addon watcher
Add a way to watch for file changes to plugins or themes (local only, not asar) and auto reload that addon.

asportnoy:
- Automatically reload addons when a change is made. This is handled automatically by the build script via RPC. Users can disable it with the `--no-reload` option.
- Moved RPC to its own coremod so it can be used between multiple coremods
- Fix an issue where plugin stylesheets would sometimes not be deleted when a plugin is stopped
- Tweak lint config

Closes #361

## Testing
1) Checkout, pull, install node modules
2) `pnpm run prepublishOnly && npm run build`
3) Reload Discord
4) In your plugin: `pnpm link /path/to/replugged`
5) Run the build or watch scripts as usual


Tharki-God:
Hi I was wondering if you can make it so it prompts to reload discord itself if plugin have plain text patches in it

asportnoy:
You might as well just ctrl+r yourself since it would require a prompt

Tharki-God:
> You might as well just ctrl+r yourself since it would require a prompt

well sometimes I just forget to restart discord (must not be only me, right? RIGHT?) thats why it would be nice if there was a prompt. maybe add args on the command to make it not give the prompt

PR #438, opened 2023-03-31
Use WeakRef for uninject functions

colin273:
Tiny change that I've been wanting to make for a while: in the injector, instead of referring directly to anything in the uninjector that gets returned from `inject()`, use `WeakRef` so that if the object has been garbage collected, no strong reference to any part of the injection is held.

Also use TypeError instead of Error if a plugin attempts to inject into a non-function


lexisother:
Please fix the check runs, then it's good to go :)

asportnoy (approved):
Haven't tested but I think this looks good

PR #449, opened 2023-04-14
feat: beginnings of context menu api

Catalyst4222:
This pr adds an api to add items to context menus, through the `injector.utils.addMenuItem` function

A few examples:
```jsx 
import { Injector, types, components } from "replugged";
const { ContextMenuTypes } = types
const { MenuItem } = components.ContextMenu;

const injector = new Injector();

export function start() {
  function getItem(data, menu) {
    return (
      <MenuItem 
        label="An item"  // What shows to the user
        id="id" 
        action={() => console.log(data)}  // What gets called when the item is clicked
      />
    );
  }

  // Registering the item
  // First arg is the navId of the menu
  // Second arg is the function  that gets called
  injector.utils.addMenuItem(ContextMenuTypes.UserContext, getItem);

  // If you don't want to use jsx, you can instead return an object
  // The attributes should correspond to the properties needed, except for `type`, which is the type of the item
  injector.utils.addMenuItem(ContextMenuTypes.Message, (data) => {
    return {
      id: "item",
      label: "Item",
      action: () => console.log(data),
      type: MenuItem,
    };
  });

  // Registering an item also returns a function to remove the item if need be
  const callToRemove = injector.utils.addMenuItem(ContextMenuTypes.ChannelContext, (data) => {


    // You can add children to make the pop-out menu, like the "Invite to Server" on a user menu
    return {
      type: MenuItem,
      label: "Hold on...",
      id: "abcd",
      children: [
        {
          type: MenuItem,
          label: "Woah a submenu!",
          id: "1234",
          action: () => console.log(data),
        },
      ],
    };

    // Same as the above, but with jsx
    <MenuItem
      label="Hold on..."
      id="abcd"
    >
      <MenuItem
        label="Woah a submenu!"
        id="1234"
        action={() => console.log(data)}
      />
    </MenuItem>
  });

  // Let's say you hate the copy-id button even existing, and want to remedy that
  injector.utils.addMenuItem(ContextMenuTypes.GuildContext, (data, menu) => {
    // Simply edit the menu, and it'll no longer appear
    menu.props.children.pop()
    
    // We don't want to show an item, so we simply return nothing
    // Anything falsy can be returned in place of an item, and discord ignores it
  })
}

export function stop() {
  injector.uninjectAll(); // Would also remove any registered items
}
```

I'm not the most experienced with typescript, so if there's anything that needs to be changed, let me know


src/renderer/coremods/contextMenu/plaintextPatches.ts
```diff
@@ -0,0 +1,23 @@
+import type { PlaintextPatch } from "src/types";
+
+export default [
+  {
+    find: "♫ (つ｡◕‿‿◕｡)つ ♪",
+    replacements: [
+      {
+        match: /(function ..?\((.)\){)(var .,.=.\.navId)/,
+        // replace: `$1(()=>{replugged.coremods.coremods.contextMenu._insertMenuItems.($2)})();$3`,
```
asportnoy, line 9:
```suggestion
```

src/renderer/coremods/contextMenu/plaintextPatches.ts
```diff
@@ -0,0 +1,23 @@
+import type { PlaintextPatch } from "src/types";
+
+export default [
+  {
+    find: "♫ (つ｡◕‿‿◕｡)つ ♪",
```
asportnoy, line 5: This property is not required. I think your query is specific enough that we don't have to worry about this.
```suggestion
```

src/renderer/coremods/contextMenu/plaintextPatches.ts
```diff
@@ -0,0 +1,23 @@
...
+    replacements: [
+      {
+        match: /(function ..?\((.)\){)(var .,.=.\.navId)/,
```
asportnoy, line 8: Use `\w+` instead of `.`
```suggestion
        match: /(function \w+\((\w+)\){)(var \w+,\w+=\w+\.navId)/,
```

src/renderer/coremods/contextMenu/plaintextPatches.ts
```diff
@@ -0,0 +1,23 @@
...
+        // replace: `$1(()=>{replugged.coremods.coremods.contextMenu._insertMenuItems.($2)})();$3`,
+        replace: `$1replugged.coremods.coremods.contextMenu._insertMenuItems($2);$3`,
+      },
+    ],
+  },
+  {
+    find: /navId:[\w"-]+,/,
```
asportnoy, line 15:
```suggestion
```

asportnoy (changes requested):
Few things I found from my first look through

src/renderer/coremods/contextMenu/plaintextPatches.ts
12944qwerty: Keeping a `find` eases the load on searching for that regex. it should be fine

src/renderer/coremods/contextMenu/index.tsx
```diff
@@ -0,0 +1,216 @@
...
+import { getByProps } from "../../modules/webpack";
+import { React } from "@common";
+
+const logger = Logger.api("ContextMenu");
+
+/**
+ * An enum for the navIds of context menus across discord
+ * @enum {string}
+ */
+export enum navIds {
```
FedeIlLeone, line 16:
```suggestion
export enum NavIds {
```
Update all other types that use the enum NavIds.

FedeIlLeone (changes requested):

src/renderer/coremods/contextMenu/index.tsx
```diff
@@ -0,0 +1,216 @@
...
+import { getByProps } from "../../modules/webpack";
+import { React } from "@common";
+
+const logger = Logger.api("ContextMenu");
+
+/**
+ * An enum for the navIds of context menus across discord
+ * @enum {string}
+ */
+export enum NavIds {
```
asportnoy, line 16: Can we rename this to something more clear? Maybe `ContextMenuTypes` or something like that. Also I think it should be moved to the types folder.

src/renderer/modules/injector.ts
```diff
@@ -1,7 +1,9 @@
 import type { ObjectExports } from "../../types/webpack";
 import type { AnyFunction } from "../../types/util";
 import type { GetButtonItem } from "../../types/coremods/message";
+import type { GetContextItem } from "../../types/coremods/contextMenu";
 import { addButton } from "../coremods/messagePopover";
+import { addContextMenuItem, NavIds } from "../coremods/contextMenu";
```
asportnoy, line 6: Lint
```suggestion
import { NavIds, addContextMenuItem } from "../coremods/contextMenu";
```

asportnoy (changes requested):
Few more minor things, then I'll test

src/renderer/modules/injector.ts
```diff
@@ -307,6 +309,39 @@ export class Injector {
...
+     * @param item The function that creates the item to add
+     * @returns A callback to de-register the function
+     *
+     * @example
+     * ```
+     * import { Injector, webpack } from "replugged";
+     * const inject = new Injector();
+     *
+     * function start() {
+     *   injector.utils.addMenuItem("user-context",  // Right-clicking a user
```
asportnoy, line 325: One more thing: fix the navId param here to use the enum

FedeIlLeone (changes requested):

src/renderer/coremods/contextMenu/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+    // If something falsy is passed, let it through
+    // Discord just skips over them too
+    return raw;
+  }
+  if (React.isValidElement(raw)) {
+    // We can't construct something that's already made
+    return raw as unknown as ContextItem;
+  }
+
+  raw = raw as RawContextItem;
```
asportnoy, line 34: ESLint says this is unnecessary, please double check if that's the case and either remove this or ignore the warning.

src/renderer/coremods/contextMenu/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  const { type, ...props } = raw;
+  if (props.children) {
+    props.children = props.children.map((child: RawContextItem | ContextItem | undefined) =>
+      makeItem(child),
+    );
+  }
+
+  return React.createElement(type, props as Record<string, unknown>) as unknown as ContextItem;
```
asportnoy, line 43: Do you need to assert unknown? Should be avoided if possible.

src/renderer/coremods/contextMenu/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * Add an item to any context menu
+ * @param navId The id of the menu you want to insert to
+ * @param getItem A function that creates and returns the menu item
+ * @returns A callback to de-register the function
+ */
+export function addContextMenuItem(navId: ContextMenuTypes, getItem: GetContextItem): () => void {
+  if (!menuItems[navId]) menuItems[navId] = [];
+
+  // Need to cast to GetContextItem since the type parameter is confusing it
```
asportnoy, line 55: My b on this one
```suggestion
```

asportnoy (commented):
Pretty good overall! Few minor requests:
- Add a random ID suffix to the ID like we have on the popover API
- Add some option to change where the item is added. Maybe a function that indicates where it should be added? I think you could do a `sectionId` option (defaults to the replugged section) and a `sectionIndex` option (defaults to the end of that section).

Tharki-God:
Can you make it so if returned value is not the type menu item or menu group but context menu itself then it replaces whole context menu? Yeah you can do it by replacing all the props manually too but this would be easier for development

asportnoy (approved):

src/renderer/modules/components/ContextMenu.tsx
```diff
@@ -1,10 +1,10 @@
 import type React from "react";
-import type { ObjectExports } from "../../../types";
+import type { ObjectExports, ReactComponent } from "../../../types";
 import { filters, getFunctionBySource, sourceStrings, waitForModule } from "../webpack";
 
 interface MenuProps {
   navId: string;
-  children: React.ReactNode;
+  children: React.ReactElement[];
```
FedeIlLeone, line 7: With that you are now forced to put multiple childrens.

Catalyst4222:
ESLint failing is unrelated to this pr

asportnoy:
Lint issues resolved

FedeIlLeone (approved):

PR #450, opened 2023-04-14
Webpack sanity

### LINKED ISSUE #440: Allow webpack module to search by props in functions
### Describe the bug

Currently, the webpack searcher only looks for props in objects, ignoring functions. This makes it needlessly verbose to search for modules such as moment.js (see #439), where useful props are on the top-level `moment()` function (the only export of its module) rather than an object.

### Reproduction steps

1. Try to search for moment.js using `replugged.webpack.getByProps("isMoment")`
2. Fail
3. Use `replugged.webpack.getModule(m => typeof m.exports === "function" && m.exports.isMoment)` to get moment.js
4. Grumble about how relatively tedious that is

### Expected behavior

`getByProps` should find moment.js

### Actual behavior

`getByProps` returns undefined

### Additional information

Should be fixed soon-ish in the `webpack-sanity` branch, whenever that makes it to the PR stage

colin273:
Make the webpack searcher code more readable and robust. No more 700-something-line file.

- Split the code into multiple files for easier maintenance and separation of tasks. Also makes overload hell less daunting, and allows us to keep more of our internals hidden from the public interface.
- Simplify the types considerably.
- Improve stability in searches by using try-catch to handle misbehaving properties (`DOMTokenList.prototype.length` for example)
- Allow props to be searched in functions as well as on objects (closes #440)


i18n/cs.json
asportnoy, line 1: Can you add this in Weblate instead (same with the other one too)

src/renderer/modules/webpack/get-modules.ts
```diff
@@ -0,0 +1,182 @@
...
+/**
+ * Get the default export (or top-level exports) for a module
+ * @param m Raw module
+ * @returns Exports
+ * @internal
+ */
+export function getExports<T>(m: RawModule): T | undefined {
+  if (typeof m.exports === "object" && m.exports) {
+    if (Object.keys(m.exports).length === 1) {
+      for (const key of ["Z", "default"] as const) {
```
lexisother, line 16: Where is `ZP`?

colin273: At the time we set these, during the original rewrite, `Z` was the most common default export key for modules built with SWC, and `default` was for stragglers that had been compiled with Babel. Is `ZP` common enough at this point to add it to this list? Currently I see:

- 13 modules with ZP as their sole export
- 154 modules with ZP as an export in general, including other top-level exports
- 1364 modules with Z as their sole export
- 1547 modules with Z as an export in general

lexisother: I think it is important enough. I have only seen Z and ZP as sole export names. Given that ZP exists as a sole export at all makes it important enough in my eyes.

asportnoy (approved):

PR #452, opened 2023-04-18
feat(badges): change module, add sizes and memoize current cache

FedeIlLeone:
- Fixed badges _again_.
- Added badge sizes.
- Memoize current cached badges (fixes [this](https://gyazo.com/1e311ff5725e6a086399cfd7076c8c36) issue).
- Updated Booster (now centered) and Support (made bigger) badge.
- Changed all SVGs to size 24 for consistency.
---
- Includes two new type functions for `users` and removed a type function from `channels` common module.


asportnoy (approved):
Works as expected

PR #456, opened 2023-05-13
feat: support replugged store

asportnoy:
Support the store backend from https://github.com/replugged-org/replugged-backend/pull/5 as an installer/updater source


src/main/ipc/installer.ts
```diff
@@ -98,11 +85,44 @@ async function github(
   };
 }
 
+const STORE_BASE_URL = `https://replugged.dev/api/v1/store`;
```
12944qwerty, line 88: this should try to use the backend URL from settings that a user can specify

asportnoy: @12944qwerty we do not have an option for they currently so that's out of scope for this PR.

src/main/ipc/installer.ts
```diff
@@ -98,11 +86,47 @@ async function github(
...
+async function store(id: string): Promise<CheckResultSuccess | CheckResultFailure> {
+  const apiUrl =
+    ((await readTransaction("dev.replugged.Settings", (settings) => settings.get("apiUrl"))) as
+      | string
+      | undefined) ?? "https://replugged.dev";
+  const STORE_BASE_URL = `${apiUrl}/api/v1/store`;
+  const manifestUrl = `${STORE_BASE_URL}/${id}`;
+  const asarUrl = `${manifestUrl}.asar`;
+
+  const res = await fetch(`${STORE_BASE_URL}/${id}`);
```
lexisother, line 98: Unneeded code duplication.
```suggestion
  const res = await fetch(manifestUrl);
```

lexisother (changes requested):

colin273 (approved):
Hope this works for people

PR #459, opened 2023-05-17
Support NodeJS v20 for scripts

asportnoy:
The build and inject scripts were not working on Node v20. The main fix was to use the `.mts` extension to force the files to be ESM (we cannot use type module due to some Electron things). Now it works on both v20 and v18.


lexisother (approved):
Works.

PR #472, opened 2023-06-04
getByPrototype

### LINKED ISSUE #375: getByPrototype
### Describe the feature request

A function to get by props on prototype 

### Why should this be added?

It should be a good addition to webpack module 

### Additional information

_No response_

- [X] This is a feature request for Replugged itself, not a feature request for a plugin or theme
- [X] [I searched the issues](https://github.com/replugged-org/replugged/issues) to make sure this feature has not been requested before.

Tharki-God:
adds a function to webpack to get function with specified prototype properties. 
returns the function itself if not getting raw/all modules.
closes #375 

note: i did the testing and it works as of function, idk about types. pls test them <3


src/renderer/modules/webpack/get-modules.ts
```diff
@@ -65,6 +65,62 @@ export function getExportsForProps<T, P extends PropertyKey = keyof T>(
...
+      // There could be other cases too, hence this try-catch instead of a specific exclusion
+      if (typeof m === "object") {
+        for (const key in m) {
+          const val = (m as Record<PropertyKey, unknown>)[key];
+          if (val && typeof val === "function") {
+            yield val as AnyFunction;
+            continue;
+          }
+          if (typeof val === "object") {
+            for (const subkey in val) {
```
asportnoy, line 89:
```suggestion
            for (const subKey in val) {
```

src/renderer/modules/webpack/get-modules.ts
```diff
@@ -65,6 +65,62 @@ export function getExportsForProps<T, P extends PropertyKey = keyof T>(
...
+              const subVal = (val as Record<PropertyKey, unknown>)[subkey];
```
asportnoy, line 90:
```suggestion
              const subVal = (val as Record<PropertyKey, unknown>)[subKey];
```

asportnoy (changes requested):
For cSpell

src/renderer/modules/webpack/get-modules.ts
```diff
@@ -65,6 +65,62 @@ export function getExportsForProps<T, P extends PropertyKey = keyof T>(
   }
 }
 
+/**
```
asportnoy, line 68: How is this different from the one we already have above? Would much rather modify that to be reusable rather than create another one which is basically the same.

Tharki-God: if you mean iterateModuleFunctions function it gets functions from the first level and second level of the object. I didn't knew how to keep this inside an already existing function while keeping it have same functionality so did this

asportnoy: Add a parameter to control which behavior it should use

Tharki-God: done

asportnoy (changes requested):
Should be renamed to `getByPrototype` (singular)

Tharki-God:
> Should be renamed to `getByPrototype` (singular)

keeping it consistent with `getByProps` it shouldn't be singular

asportnoy:
> > Should be renamed to `getByPrototype` (singular)
> 
> keeping it consistent with `getByProps` it shouldn't be singular

Props is plural because there are multiple props that you're getting by. There's only one prototype. Plus, `getBySource` is also singular.

colin273:
> keeping it consistent with `getByProps` it shouldn't be singular
What? No. Singular is more appropriate here. If you're just getting one prototype with the filter, there is no scenario in which plural makes sense. Use what fits the meaning.

Tharki-God:
> > > Should be renamed to `getByPrototype` (singular)
> > 
> > 
> > keeping it consistent with `getByProps` it shouldn't be singular
> 
> Props is plural because there are multiple props that you're getting by. There's only one prototype. Plus, `getBySource` is also singular.

nope multiple prototypes are used to get the function usually for good measures. it works just like getByProps

Tharki-God:
this is how its used ``webpack.getByPrototypes(["updateVideoQuality", "setUseElectronVideo"], {raw:false,all:false})``

asportnoy (approved):

PR #479, opened 2023-06-11
Permission fix and other something

momu54:
+ Fix permission of the config folder
+ Fix Linux dependency install issue
+ Remove unnecessary util call

- [x] All changes tested

### Distro Tasted

- [x] Ubuntu based
- [x] Arch based

PR #482, opened 2023-06-13
Apply permission fix on macOS

momu54:
From #479 

Apply that fix on macOS (Needs macOS user test)

- [ ] macOS tested


momu54:
I googled permission of macOS and I see this patch isn't in needed
I'm so sorry about that 
Thank every maintainers

momu54, 2023/6/13

Issue #383, opened 2023-02-12
Nesed setting support 


Tharki-God:
### Describe the feature request

support for nested settings for example in my plugin the settings are   ```
fps: {
    1: 15,
    2: 30,
    3: 60,
  },
  resolution: {
    1: 480,
    2: 720,
    3: 1080,
  },
  smoothVideo: {
    resolution: 720,
    fps: 60,
  },
  betterReadability: {
    resolution: 0,
    fps: 60,
  },
```

and i cant use util.useSettings with this because it want the setting to be direct instead of nested object

### Why should this be added?

for cleaner settings it would be nice

### Additional information

_No response_

### Please confirm the following

- [X] This is a feature request for Replugged itself, not a feature request for a plugin or theme
- [X] [I searched the issues](https://github.com/replugged-org/replugged/issues) to make sure this feature has not been requested before.

PR #487, opened 2023-06-18
util update

Tharki-God:
added support for nested settings and fixed type for findInReactTree

this pr will close https://github.com/replugged-org/replugged/issues/383

sorry closed the old pr by mistake coz of renaming the branch. sorry

### DIFF
--- a/src/renderer/util.ts
+++ b/src/renderer/util.ts
@@ -1,5 +1,6 @@
-import { React, channels, fluxDispatcher, guilds } from "@common";
+import { React, channels, fluxDispatcher, guilds, lodash } from "@common";
 import type { Fiber } from "react-reconciler";
+import type { ReactElement } from "react";
 import type { Jsonifiable } from "type-fest";
 import type { ObjectExports } from "../types";
 import { SettingsManager } from "./apis/settings";
@@ -182,7 +183,7 @@ export function useSetting<
   F extends T[K] | undefined,
 >(
   settings: SettingsManager<T, D>,
-  key: K,
+  key: keyof T & string extends K ? string : K,
   fallback?: F,
 ): {
   value: K extends D
@@ -192,8 +193,10 @@ export function useSetting<
     : NonNullable<T[K]> | F;
   onChange: (newValue: ValType<T[K]>) => void;
 } {
-  const initial = settings.get(key, fallback);
-  const [value, setValue] = React.useState(initial);
+  const [initialKey, ...pathArray] = key.split(".") as K[];
+  const path = pathArray.join(".");
+  const initial = settings.get(initialKey, fallback);
+  const [value, setValue] = React.useState(path.length ? lodash.get(initial, path) : initial);
 
   return {
     value,
@@ -209,9 +212,11 @@ export function useSetting<
       const targetChecked = target && "checked" in target ? target.checked : undefined;
       const finalValue = (checked ?? targetChecked ?? targetValue ?? value ?? newValue) as T[K];
 
-      // @ts-expect-error dumb
       setValue(finalValue);
-      settings.set(key, finalValue);
+      settings.set(
+        key as K,
+        path.length ? (lodash.set(initial as object, path, finalValue) as T[K]) : finalValue,
+      );
     },
   };
 }
@@ -353,11 +358,11 @@ export function findInTree(
  * @returns The component you are looking for
  */
 export function findInReactTree(
-  tree: Tree,
+  tree: Tree | ReactElement,
   searchFilter: TreeFilter,
   maxRecursion = 100,
-): Tree | null | undefined {
-  return findInTree(tree, searchFilter, {
+): Tree | ReactElement | null | undefined {
+  return findInTree(tree as Tree, searchFilter, {
     walkable: ["props", "children", "child", "sibling"],
     maxRecursion,
   });


asportnoy (changes requested):
To prevent breaking changes, add a fallback as the original string with the dot included.

Tharki-God:
> To prevent breaking changes, add a fallback as the original string with the dot included.

i dont get what you mean. if you think not including . will break for not nested settings then its not the case ig

asportnoy:
If an addon is already relying on a key like `hello.world`, that data essentially will be wiped after this PR. So the initial behavior needs to be added as a fallback on the getter.

Tharki-God:
How do I add that? Any ideas?

asportnoy:
Just add a fallback to use the initial behavior if the new behavior is undefined

Tharki-God:
How do I decide if new behavior is undefined or not. What is the new behavior anyways

Tharki-God:
fixed types, and fallback value usage. also added a check if the key given is already in settings ignoring the path separator it will be treated as a key. 
well should be enough for what you wanted

src/renderer/util.ts
```diff
@@ -183,29 +184,42 @@ type ValType<T> =
...
   fallback?: F,
 ): {
-  value: K extends D
-    ? NonNullable<T[K]>
-    : F extends null | undefined
-    ? T[K] | undefined
-    : NonNullable<T[K]> | F;
-  onChange: (newValue: ValType<T[K]>) => void;
+  value: NonNullable<NestedType<T, P>>;
+  onChange: (newValue: ValType<NonNullable<NestedType<T, P>>>) => void;
```
asportnoy, line 207: You can't just wrap the whole thing in `NonNullable`, it needs to match how it was done before.

src/renderer/util.ts
```diff
@@ -183,29 +184,42 @@ type ValType<T> =
...
+  const [initialKey, ...pathArray] = Object.keys(settings.all()).includes(key)
+    ? ([key] as [K])
+    : (key.split(".") as [K, ...string[]]);
+  const path = pathArray.join(".");
+  const initial = settings.get(initialKey, path.length ? ({} as T[K]) : (fallback as T[K]));
+  const [value, setValue] = React.useState<NonNullable<NestedType<T, P>>>(
+    path.length
+      ? (lodash.get(initial, path, fallback) as NonNullable<NestedType<T, P>>)
+      : (initial as NonNullable<NestedType<T, P>>),
+  );
```
asportnoy, line 218: This is all way more complicated than it needs to be. Just do something like `settings.get(key) ?? lodash.get(settings.all(), key) ?? defaultValue)`.

src/renderer/util.ts
```diff
@@ -215,31 +229,28 @@ export function useSetting<
...
   fallback?: F,
-): [
-  K extends D
-    ? NonNullable<T[K]>
-    : F extends null | undefined
-    ? T[K] | undefined
-    : NonNullable<T[K]> | F,
-  (newValue: ValType<T[K]>) => void,
-] {
+): [NonNullable<NestedType<T, P>>, (newValue: ValType<NonNullable<NestedType<T, P>>>) => void] {
```
asportnoy, line 253: Same type issue as above

src/renderer/util.ts
```diff
@@ -361,11 +372,11 @@ export function findInTree(
...
+  tree: Tree | ReactElement,
   searchFilter: TreeFilter,
   maxRecursion = 100,
-): Tree | null | undefined {
-  return findInTree(tree, searchFilter, {
+): Tree | ReactElement | null | undefined {
+  return findInTree(tree as Tree, searchFilter, {
     walkable: ["props", "children", "child", "sibling"],
     maxRecursion,
   });
```
asportnoy, line 382: Unrelated change, does not belong in this PR

asportnoy (changes requested):

bin.mjs
lexisother, line 1: Can you remove this?

src/renderer/util.ts
```diff
@@ -85,7 +85,6 @@ export function forceUpdateElement(selector: string, all = false): void {
     all ? [...document.querySelectorAll(selector)] : [document.querySelector(selector)]
   ).filter(Boolean) as Element[];
 
-  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- May not actually have forceUpdate
```
lexisother, line 88: Why was this removed? Now there's an ESLint warning.

src/renderer/util.ts
```diff
@@ -172,41 +171,56 @@ export async function goToOrJoinServer(invite: string): Promise<void> {
 }
 
 export async function openExternal(url: string): Promise<void> {
-  const mod = getBySource<(url: string) => Promise<void>>(/href=\w,\w\.target="_blank"/);
+  const mod = getBySource<(url: string) => Promise<void>>('.target="_blank"');
   if (!mod) {
     throw new Error("Could not find openExternal");
   }
-  await mod(url);
+  return await mod(url);
```
lexisother, line 178: Why did you add this? `mod` returns `void`, adding `return` here is pointless.

lexisother (changes requested):

src/renderer/util.ts
YofukashiNo: by mistake ig, i forgor

fedeericodl (approved):
I've added type safety for nested settings.
LGTM

lexisother:
Could you give some usage examples? e.g. with `src/types/settings.ts`, how would it look like if `autoApplyQuickCss` became `quickCss.autoApply` and `addonEmbeds` became `qol.addonEmbeds`?

I understand the `useSetting` call would become `useSetting(generalSettings, 'qol.addonEmbeds')` but what about that interface and the `defaultSettings` object? Would you do it like this?
```ts
export type GeneralSettings = {
  'quickCss.autoApply'?: boolean;
  'qol.addonEmbeds'?: boolean;
};

export const defaultSettings = {
  'quickCss.autoApply': false,
  'qol.addonEmbeds': true,
} satisfies Partial<GeneralSettings>;
```

lexisother:
Implementing the example I've mentioned just now yields the following result. Why is `qol` not being put into an object?

image

yofukashino:
> Could you give some usage examples? e.g. with `src/types/settings.ts`, how would it look like if `autoApplyQuickCss` became `quickCss.autoApply` and `addonEmbeds` became `qol.addonEmbeds`?
> 
> I understand the `useSetting` call would become `useSetting(generalSettings, 'qol.addonEmbeds')` but what about that interface and the `defaultSettings` object? Would you do it like this?
> 
> ```ts
> export type GeneralSettings = {
>   'quickCss.autoApply'?: boolean;
>   'qol.addonEmbeds'?: boolean;
> };
> 
> export const defaultSettings = {
>   'quickCss.autoApply': false,
>   'qol.addonEmbeds': true,
> } satisfies Partial<GeneralSettings>;
> ```

Probably like a normal object. 

Example from show hidden channels

```ts
  export interface Settings {
    hiddenChannelIcon: string;
    customHiddenChannelIcon: string;
    faded: boolean;
    sort: string;
    showPerms: boolean;
    showAdmin: string;
    stopMarkingUnread: boolean;
    shouldShowEmptyCategory: boolean;
    alwaysCollapse: boolean;
    debugMode: boolean;
    channels: {
      GUILD_TEXT: boolean;
      GUILD_VOICE: boolean;
      GUILD_ANNOUNCEMENT: boolean;
      GUILD_STORE: boolean;
      GUILD_STAGE_VOICE: boolean;
      GUILD_FORUM: boolean;
    };
    blacklistedGuilds: Record<string, boolean>;
    collapsed: Record<string, boolean>;
  }
}
    ```
    
    Used like 
    ```ts
      {Object.keys(SettingValues.get("channels", defaultSettings.channels)).map(
          (type: keyof typeof defaultSettings.channels) => (
            <SwitchItem
              key={type}
              {...Utils.useSetting(
                SettingValues,
                `channels.${type}`,
                defaultSettings.channels[type],
              )}>
              {`Show ${Utils.capitalizeFirst(type.split("_")[1])}${
                type.split("_").length === 3 ? ` ${Utils.capitalizeFirst(type.split("_")[2])}` : ""
              } Channels`}
            </SwitchItem>
          ),
        )}```

yofukashino:
> Mind you that `generalSettings.get('qol').addonEmbeds` _does_ work. That's fine in my book, but I'm wondering if you want to adjust the implementation there too.


This is just meant for hooks as reading like this aint that big of mess but the custom hooks for this are

lexisother:
Would you be opposed to adapting the `SettingsManager` as well? If you're fetching a single setting it'd be nice to just give it `qol.addonEmbeds` instead of `.get('qol').addonEmbeds`.

yofukashino:
> Would you be opposed to adapting the `SettingsManager` as well? If you're fetching a single setting it'd be nice to just give it `qol.addonEmbeds` as an argument instead of having to do `.get('qol').addonEmbeds`.

I wouldn't be opposed, but I don't think it's required as you dont need to write something extra just the position.

lexisother:
I would prefer it be added in this PR. I'm of the opinion that if `useSetting` supports it then so should the rest of the API.

yofukashino:
> I would prefer it be added in this PR. I'm of the opinion that if `useSetting` supports it then so should the rest of the API.

Same for set?  and has? imo its required only in the hook

lexisother (approved):

Issue #400, opened 2023-02-26
Injection List


Tharki-God:
### Describe the feature request

it would be nice if the patched function have a list specifically an set of string with names/id of plugins it have been patched from. 

### Why should this be added?

to make it easy for plugin devlopers to check if an function have already been patched 

### Please confirm the following

- [X] This is a feature request for Replugged itself, not a feature request for a plugin or theme
- [X] [I searched the issues](https://github.com/replugged-org/replugged/issues) to make sure this feature has not been requested before.

FedeIlLeone:
We have decided not to include this anymore. Refer to https://github.com/replugged-org/replugged/pull/491#issuecomment-1599756144.

PR #491, opened 2023-06-20
Added Injection list

Tharki-God:
added a list for injections done on function, will close https://github.com/replugged-org/replugged/issues/400

also adds the original function as a value on the patched function in case the original function is needed


colin273:
The lack of an injection list is not an oversight/bug/accident, but a conscious design choice that I made when writing this injector module.

If you need to see whether a function has been patched and by whom…why? Surely that's a sign of some greater code design problem if you are depending on other plugins to either do or not do something.

The original function is already accessible outside of the injection, and for `instead` patches inside the injection as well. I don't understand why this part is even remotely necessary.

Tharki-God:
> The lack of an injection list is not an oversight/bug/accident, but a conscious design choice that I made when writing this injector module.
> 
> If you need to see whether a function has been patched and by whom…why? Surely that's a sign of some greater code design problem if you are depending on other plugins to either do or not do something.
> 
> The original function is already accessible outside of the injection, and for `instead` patches inside the injection as well. I don't understand why this part is even remotely necessary.

this injection list is for plugins that do nested patching to check if the function is already patched or not. if yes then by my patch or some other patch like that. 
also outside the patch, you can access the original function with func.prototype.constructor which is kinda a mess so for cleaner code it's better to have it directly on the function instead of whatever prototype method.

PR #495, opened 2023-06-25
context menu api fix

Tharki-God:
for some people, context menu API isn't working including myself. this is a fix as it wouldn't break if it's working for you but fix for others. also makes the plain text patch smaller and more reliable.

added find as Moongirl  was complaining about the lack of them in the replugged-dev channel


lexisother:
While you're at it, please move this way from dollar sign notation and use a function instead.

asportnoy (approved):

PR #500, opened 2023-06-27
fix(main): Fix `electron-devtools-installer` & Watch Together

GriefMoDz:
... again

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -119,6 +119,7 @@ electron.app.once("ready", () => {
       return done({});
     }
 
+    const hasFrameOptions = Object.keys(responseHeaders).find((e) => /x-frame-options/i.test(e));
     const hasAllowCredentials = Object.keys(responseHeaders).find((e) =>
       /access-control-allow-credentials/i.test(e),
     );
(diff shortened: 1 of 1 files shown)


lexisother (changes requested):
Not a real request for review, just blocking.

@CanadaHonk Please review.

CanadaHonk:
Why is a custom CSP with `frame-ancestors` being set at all? It just limits what origins can be an iframe in the page, which I don't see any reason to do?

GriefMoDz:
> Why is a custom CSP with `frame-ancestors` being set at all? It just limits what origins can be an iframe in the page, which I don't see any reason to do?

It actually does the opposite. By stripping the `x-frame-options` header and using `frame-ancestors` in its place, it opens up the use of embedded content.

As a result, plugins can now use embedded content (`iframe`s and other surrounding elements) on a majority of websites, as long as the request originates from within Discord's domain(s) - as specified under the `frame-ancestors` policy.

This is particularly useful for plugins that wish to embed a website such as "https://duckduckgo.com" that would previously fail due to it being blocked because it hasn't come from the same origin.

CanadaHonk:
Oh lol I got it confused with `frame-src`. I agree, lgtm.

CanadaHonk (approved):

lexisother (approved):

PR #502, opened 2023-06-29
Add timeout

### LINKED ISSUE #473: Add timeout for plugin ignition
Currently, a plugin can be loading forever and will block the ignition finishing. We should add a timeout for a plugin loading (maybe 5s), at which point we'll log an error and continue with the ignition. If the plugin loads later then great, but the ignition should be unblocked by that.

bop34:
Closes #473


src/renderer/managers/plugins.ts
```diff
@@ -79,7 +79,13 @@ export async function start(id: string): Promise<void> {
       plugin.exports = await import(
         `replugged://plugin/${plugin.path}/${plugin.manifest.renderer}?t=${Date.now()}}`
       );
-      await plugin.exports!.start?.();
+
+      await Promise.race([
+        new Promise((_, reject) =>
+          setTimeout(() => reject(new Error("Plugin took too long to start")), 5000),
```
asportnoy, line 85: Nit
```suggestion
          setTimeout(() => reject(new Error("Plugin took too long to start")), 5_000),
```

asportnoy (commented):
Need to test this later but one quick thing first

asportnoy (approved):

PR #504, opened 2023-06-30
feat: Use production by default

momu54:
This pr make the injector use production mode by default.

Because the new user want to install Replugged on Debian based Linux
(electron injector can't use on Debian package, because permission)
they don't know that they need to add the `--production`arg to plug command
So I made this change!


lexisother:
What's the difference between `true` and `!false`?

momu54:
I think I don't need to change other code after that

lexisother:
What's the difference between `includes('--production')` and `!includes('--development')`?

momu54:
Look at title of the pr

lexisother:
You think I didn't? harold

Anyways, I just had to work out the logic in my head. My stance remains the same.

> they don't know that they need to add the `--production` arg to plug command

That's their problem. Users should learn to read the very detailed instructions we give them. Seriously, it's written right on the damn installation page: https://replugged.dev/download

momu54:
Generally, production should be set by default
(Like any other software)
I will open this pr anyways.
If any member close it, I have no opinion.

lexisother:
Sorry, but we've decided not to integrate this.

If you want to manually inject for a prod installation, which is unlikely to happen normally, only if a user is experiencing issues with our installers, you read the instructions which includes the flag.

In any other case where you are manually injecting it is safe to assume it's for development purposes.

PR #507, opened 2023-07-03
byPrototype

yofukashino:
adding an option to getByProps to search only in the prototype, like requested before closing last pr.


src/renderer/modules/webpack/get-modules.ts
```diff
@@ -22,44 +22,66 @@ export function getExports<T>(m: RawModule): T | undefined {
 }
 
 /**
- * Iterates over an object and its top-level children that could have properties
+ * Iterates over an object and its top-level and second-level (if specified) children that could have properties
```
asportnoy, line 25: The option should be to make this check in the prototype, not to check a second a second level.

src/renderer/modules/webpack/helpers.ts
asportnoy, line 1: I think `prototype` is a better option name than `byPrototype`. Also, you made a type interface for the options, please use it consistently.

src/types/webpack.ts
```diff
@@ -73,3 +73,8 @@ export interface WaitForOptions {
   /** If nothing is found after this delay (ms), stop and throw an error. */
   timeout?: number;
 }
+
+export interface byPropsOptions {
```
asportnoy, line 77: Interfaces should always be PascalCase

src/renderer/modules/webpack/helpers.ts
```diff
@@ -246,7 +248,7 @@ export function getByValue<T>(
   return getModule<T>(filters.byValue(match), options);
 }
 
-export function getByStoreName<T extends Store>(name: string): T | undefined {
+export function getByStoreName<T>(name: string): (T & Store) | undefined {
   const stores = Flux.Store.getAll();
-  return stores.find((store) => store.getName() === name) as T | undefined;
+  return stores.find((store) => store.getName() === name) as (T & Store) | undefined;
 }
```
asportnoy, line 254: Seems like an unrelated change

asportnoy (changes requested):

src/renderer/modules/webpack/get-modules.ts
Tharki-God: okay so that's handled in filter because if we get prototype here then we wouldn't be able to return the parent function for those prototype

src/renderer/modules/webpack/helpers.ts
Tharki-God: yeah was frustrated on that so fixed it kinda. was in webpack too so thought would be fine in this update. this will make it so generic wouldn't have to extend store

src/types/webpack.ts
Tharki-God: okay will fix that

src/renderer/modules/webpack/helpers.ts
Tharki-God: I mean it's getByProps so byPrototype seems like better fit tbh and will use the interface.

src/renderer/modules/webpack/get-modules.ts
asportnoy: I'm not saying to return the prototype, I'm saying it should only be checking prototype specifically, not just any second-level object

src/renderer/modules/webpack/helpers.ts
asportnoy: That should not be in this PR though.

asportnoy (dismissed):
Additional changes above

src/renderer/modules/webpack/get-modules.ts
Tharki-God: that would need us to pass the props to check here too. while we can already do that in the function where this is used. and even if we add check here this would still be iterator and will add just some more milliseconds imo

asportnoy: No you wouldn't

Tharki-God: well, it already returns only objects and functions. only those can have prototypes anyway so idk what the issue is, I just read the code after some time, I forgot it already did this.

Tharki-God: https://github.com/Tharki-God/replugged/blob/byPrototype/src/renderer/modules/webpack/get-modules.ts#L48

PR #509, opened 2023-07-06
feat: add React DevTools

FedeIlLeone:
- Setting to enable React DevTools.
- Advanced settings have been moved to a new dedicated category.

Closes: #253


src/main/ipc/react-devtools.ts
```diff
@@ -0,0 +1,30 @@
...
+import fetch from "node-fetch";
+import { join } from "path";
+import { CONFIG_PATHS } from "src/util.mjs";
+import { RepluggedIpcChannels } from "../../types";
+
+const OUTPUT_PATH = join(CONFIG_PATHS["react-devtools"]);
+
+// https://github.com/facebook/react/issues/25843#issuecomment-1406766561
+const EXTENSION_URL =
+  "https://github.com/mondaychen/react/raw/017f120369d80a21c0e122106bd7ca1faa48b8ee/packages/react-devtools-extensions/ReactDevTools.zip";
```
asportnoy, line 12: Per internal discussion, we want to move this to some source we control, either forking this to our GH org or hosting it directly on the server.

src/main/ipc/react-devtools.ts
```diff
@@ -0,0 +1,30 @@
...
+
+ipcMain.handle(RepluggedIpcChannels.DOWNLOAD_REACT_DEVTOOLS, async () => {
+  const arrayBuffer = await fetch(EXTENSION_URL)
```
asportnoy, line 15: Can we check if it's already there before downloading it again? Also one we should consider (which ig could be done later) is figure out how to handle updates to RDT. Maybe a setting key for the version and if that's older than whatever version is in a certain const then we re-download when Discord is launched?

src/renderer/coremods/settings/pages/General.tsx
```diff
@@ -86,42 +80,84 @@ export const General = (): React.ReactElement => {
...
+              .confirm({
+                title: Messages.REPLUGGED_SETTINGS_REACT_DEVTOOLS,
+                body: Messages.REPLUGGED_SETTINGS_REACT_DEVTOOLS_DESC.format(),
+                confirmText: Messages.REPLUGGED_RELOAD,
+                confirmColor: Button.Colors.RED,
+              })
+              .then((answer) => {
+                if (answer) {
+                  if (value) {
+                    void RepluggedNative.reactDevTools.downloadExtension().then(() => {
```
asportnoy, line 137: Add error handling for download failed. Maybe just a toast?

src/renderer/coremods/settings/pages/General.tsx
```diff
@@ -86,42 +80,84 @@ export const General = (): React.ReactElement => {
...
+                      rdtOnChange(value);
+                      setTimeout(() => window.DiscordNative.app.relaunch(), 250);
```
asportnoy, line 139: Let's create functions here called reload and relaunch (with the timeout built in) and then replace the calls in this file.

src/types/settings.ts
```diff
@@ -12,6 +12,7 @@ export type GeneralSettings = {
   badges: boolean;
   autoApplyQuickCss: boolean;
   showWelcomeNoticeOnOpen: boolean;
+  reactDevTools: boolean;
 };
 
 export const defaultSettings: Partial<GeneralSettings> = {
```
asportnoy, line 18: Unrelated to this PR but the type here should be using `satisfies`:

```ts
export const defaultSettings = {
  badges: true,
  autoApplyQuickCss: false,
  showWelcomeNoticeOnOpen: true,
  reactDevTools: false,
} satisfies Partial<GeneralSettings>;
```
GH won't let me suggest here but can you fix it in this branch?

asportnoy (changes requested):

src/main/ipc/react-devtools.ts
FedeIlLeone: > Can we check if it's already there before downloading it again?
What if the extension breaks, I don't know, the user deletes some file or changes something making it not loading. A redownload would overwrite any files, solving this possible issue. How would I check if the extension is downloaded or not? Should I save the zip individually, and check if it's there?

> Also one we should consider (which ig could be done later) is figure out how to handle updates to RDT. Maybe a setting key for the version and if that's older than whatever version is in a certain const then we re-download when Discord is launched?
Certainly later, if we will consider forking the modified RDT.

src/main/ipc/react-devtools.ts
```diff
@@ -1,23 +1,36 @@
...
-  const arrayBuffer = await fetch(EXTENSION_URL)
-    .then((res) => res.arrayBuffer())
-    .catch(() => {
-      throw new Error("Could not download React DevTools");
-    });
-  const buffer = Buffer.from(new Uint8Array(arrayBuffer));
+  const apiUrl =
+    ((await readTransaction("dev.replugged.Settings", (settings) => settings.get("apiUrl"))) as
+      | string
+      | undefined) ?? "https://replugged.dev";
```
asportnoy, line 17: Can we set up some ipc util with this? Since we're using it in store too

src/renderer/coremods/settings/pages/General.tsx
```diff
@@ -134,12 +142,21 @@ export const General = (): React.ReactElement => {
...
-                      setTimeout(() => window.DiscordNative.app.relaunch(), 250);
-                    });
+                    void RepluggedNative.reactDevTools
+                      .downloadExtension()
+                      .then(() => {
+                        rdtOnChange(value);
+                        relaunch();
+                      })
+                      .catch(() => {
+                        void toast.toast(
```
asportnoy, line 152: Not a promise so `void` shouldn't be needed
```suggestion
                        toast.toast(
```

src/renderer/coremods/settings/pages/General.tsx
```diff
@@ -86,42 +88,93 @@ export const General = (): React.ReactElement => {
...
+        <SwitchItem
+          value={rdtValue}
+          onChange={(value) => {
+            void modal
+              .confirm({
+                title: Messages.REPLUGGED_SETTINGS_REACT_DEVTOOLS,
+                body: Messages.REPLUGGED_SETTINGS_REACT_DEVTOOLS_DESC.format(),
+                confirmText: Messages.REPLUGGED_RELOAD,
+                confirmColor: Button.Colors.RED,
+              })
```
asportnoy, line 141: Can we download first and then show the modal?

asportnoy (changes requested):

lexisother:
Is this done? If there are no more comments to be made I am dismissing your review as stale. @asportnoy

lexisother (approved):

src/renderer/coremods/settings/pages/General.tsx
```diff
@@ -110,56 +124,31 @@ export const General = (): React.ReactElement => {
...
-                confirmColor: Button.Colors.RED,
-              })
-              .then((answer) => {
-                if (answer) {
-                  expOnChange(value);
-                  reload();
-                }
-              });
-          }}
+          onChange={(value) => restartModal(false, () => expOnChange(value))}
```
asportnoy, line 127: This should change the setting even if you choose to not restart yet

src/renderer/coremods/settings/pages/General.tsx
```diff
@@ -110,56 +124,31 @@ export const General = (): React.ReactElement => {
...
-                    rdtOnChange(value);
-                    relaunch();
-                  }
-                }
-              });
+            if (value) {
+              void RepluggedNative.reactDevTools
+                .downloadExtension()
+                .then(() => {
+                  rdtOnChange(value);
```
asportnoy, line 139: Do the onChange before downloading so the switch doesn't flicker, but revert the change in the catch block.

asportnoy (changes requested):
Last two things hopefully

asportnoy (approved):
image
Will merge once Fede gives the ok

PR #515, opened 2023-07-09
Addon embeds

### LINKED ISSUE #252: Plugin Embeds Coremod

asportnoy:
Screenshot on 2023-07-09 at 18 44 47

Enabled by default, but can be disabled in settings. Only store addons are supported.

Closes #252


src/renderer/coremods/installer/AddonEmbed.tsx
```diff
@@ -0,0 +1,221 @@
...
+      const info = await getInfo(addon.identifier, addon.source, addon.id);
+      if (!info) return;
+      setData(info);
+    })();
+  }, []);
+  const [onCooldown, setOnCooldown] = React.useState(false);
+  const [cooldownTimeout, setCooldownTimeout] = React.useState<NodeJS.Timeout | null>(null);
+  const [isInstalling, setIsInstalling] = React.useState(false);
+
+  if (!data) return null;
```
FedeIlLeone, line 121: Maybe, instead of returning null, return a loading embed, just like they do with the build override embed, while it's fetching the info.

src/renderer/coremods/installer/index.tsx
```diff
@@ -108,18 +123,28 @@ async function injectLinks(): Promise<void> {
     return;
   }
 
-  injector.before(exports, key, ([args]) => {
-    const { href } = args;
-    if (!href) return undefined;
+  injector.instead(exports, key, (args, fn) => {
```
FedeIlLeone, line 126: Editing the message doesn't update the embed https://gyazo.com/76026f51ab588bbb7ea7ae71a0442a99
Maybe we can also remove the classic link embed?

src/renderer/coremods/installer/AddonEmbed.tsx
```diff
@@ -0,0 +1,221 @@
...
+    setOnCooldown(true);
+    if (cooldownTimeout) clearTimeout(cooldownTimeout);
+    setCooldownTimeout(
+      setTimeout(() => {
+        setOnCooldown(false);
+      }, 2000),
+    );
+  };
+
+  return (
```
FedeIlLeone, line 163: Space is not optimized well, compared to the build override embed.
image

FedeIlLeone (changes requested):

src/renderer/coremods/installer/AddonEmbed.tsx
asportnoy: Returning null will make it return a regular link, though I just noticed an issue with that which I will fix.

lexisother: > Returning null will make it return a regular link

Yes, but why not make it return a loading embed? Back in my very first implementation of the functionality, we took this approach: https://github.com/Cumcord/Cumcord/blob/f98494df6e7ee781163a7f2a54f3197e92bf73a4/src/api/utils/useFetchModule.js

It'd then be used as a sort of "hook"-like object, which will cause the component to automatically rerender whenever the data from that hook changed: https://github.com/Cumcord/Cumcord/blob/f98494df6e7ee781163a7f2a54f3197e92bf73a4/src/ui/pluginEmbeds/components/PluginEmbed.jsx#L42

src/renderer/coremods/installer/AddonEmbed.tsx
FedeIlLeone: This is somewhat still an issue, can we stretch the install button all the way to the end (if there's not enough space) just like the apply button in the screenshot above?

src/renderer/coremods/installer/AddonEmbed.tsx
```diff
@@ -0,0 +1,295 @@
...
+      buildDetails,
+      subHead,
+      copyLink,
+      copyLinkIcon,
+      copied,
+    } = classMod;
+
+    return (
+      <div className={wrapper}>
+        <Text.Eyebrow className={titleRegion}>
```
FedeIlLeone, line 149:
```suggestion
        <Text className={titleRegion} variant="text-xs/normal">
```

FedeIlLeone (changes requested):

src/renderer/coremods/installer/AddonEmbed.tsx
asportnoy: Why? The real one uses eyebrow.

FedeIlLeone (approved):
👍

PR #517, opened 2023-07-16
Auto relaunch

yofukashino:
automatically close discord and open it back on plug/unplug/replug.

the behavior can be avoided by passing the argument `--no-relaunch`.


scripts/inject/index.mts
```diff
@@ -121,7 +142,22 @@ To plug into a different platform, use the following syntax: ${AnsiEscapes.BOLD}
     }
   } else if (cmd === "uninject") {
     try {
-      result = await uninject(platformModule, platform);
+      if (noRelaunch) {
```
asportnoy, line 145: This should be moved to a function instead of repeated twice

asportnoy (changes requested):

scripts/inject/util.mts
```diff
@@ -29,3 +30,56 @@ export const getCommand = ({
...
+  } catch {
+    return false;
+  }
+};
+
+export const killProcessByPID = (pid: number): Promise<void> => {
+  return new Promise((resolve) => {
+    if (!pid) resolve();
+    process.kill(pid, "SIGTERM");
+    const checkInterval = setInterval(() => {
```
asportnoy, line 74: Add some timeout here in case it never quits so it won't hang indefinitely

asportnoy (changes requested):

scripts/inject/injector.mts
```diff
@@ -230,3 +238,58 @@ export const uninject = async (
...
+            ...GetUserData(),
+            detached: true,
+            stdio: "ignore",
+          });
+          break;
+        case "darwin":
+          openProcess(join(appDir, "..", "..", "MacOS", processName), [], {
+            detached: true,
+            stdio: "ignore",
+          });
```
asportnoy, line 289: Instead of this, run `open -a "Discord Canary"` (or other platform name).

asportnoy (changes requested):

scripts/inject/injector.mts
asportnoy: exec

asportnoy (changes requested):
Requested changes have not been made

Tharki-God:
also made it so that if Discord wasn't open before then don't launch it after.

FedeIlLeone (approved):

colin273 (approved):

PR #519, opened 2023-07-19
Fix QuickCSS permissions on Linux and remove elevation from injector

EastArctica:
Fixed an issue where ~/.config/replugged/quickcss/main.css would be owned by root and therefore  not be editable by the discord client.

Removed the global elevation of the injector, this fixes a multitude of issues and a considerable amount of code that was specifically added to deal with the current user being root should now be able to be removed.


EastArctica:
One thing I would like added before this is merged, the user should be notified and asked about the root permissions being used in the event that they had previously authorized their session for root permissions. (ex. had used sudo in the past few minutes in their shell)

SomeAspy (approved):
Seems to work™️

src/util.mts
```diff
@@ -1,6 +1,7 @@
 import { execSync } from "child_process";
-import { chownSync, existsSync, mkdirSync, statSync, writeFileSync } from "fs";
+import { existsSync, mkdirSync, statSync, writeFileSync } from "fs";
 import { join } from "path";
+import { AnsiEscapes, PRIV_CMD_EXEC } from "../scripts/inject/util.mjs";
```
asportnoy, line 4: Can you move these things into this file and import them in the injector?

EastArctica: You want AnsiEscapes and PRIV_CMD_EXEC inside of src/util.mts? They're only used in 1 location in the src/util.ts but are used many times in the scripts/inject/util.mjs but if you'd like I can. Or alternatively, did you want it in both scripts/inject/util.mjs and src/util.mts?

asportnoy: > You want AnsiEscapes and PRIV_CMD_EXEC inside of src/util.mts?
Yes. From a structure POV, I think it's better to not be importing from scripts.

scripts/inject/injector.mts
```diff
@@ -190,20 +197,19 @@ export const inject = async (
...
+      '"',
+      '\\"',
+    )})" | ${PRIV_CMD_EXEC} tee ${join(appDir, "index.js")}`,
+  );
+  execSync(
+    `echo "${JSON.stringify({
+      main: "index.js",
+      name: "discord",
+    }).replaceAll('"', '\\"')}" | ${PRIV_CMD_EXEC} tee ${join(appDir, "package.json")}`,
+  );
```
lexisother, line 212: This will also run on Windows. Obviously won't work.

lexisother (changes requested):
@EastArctica Did you test the Injector on Windows? You replaced a couple of platform agnostic lines with literal executions of Linux-only commands without guarding them off to only be ran on Linux.

asportnoy:
Closing in favor of #533

PR #520, opened 2023-07-21
context menu api fix

yofukashino:
makes it so you can use React.useState with the api for updating your menu items.


src/renderer/coremods/contextMenu/index.tsx
```diff
@@ -88,53 +91,69 @@ type ContextMenuData = ContextMenuProps["ContextMenu"] & {
...
-
-  // Already inserted items
-  // If this isn't here, another group of items is added every update
-  if (menu.plugged) return;
-
-  // We delay getting the items until now, as importing at the start of the file causes Discord to hang
-  // Using `await import(...)` is undesirable because the new items will only appear once the menu is interacted with
-  const { MenuGroup } = getByProps<Record<string, React.ComponentType>>([
+async function injectMenu(): Promise<void> {
+  const RepluggedComponents = await waitForModule(filters.byProps("Menu", "MenuGroup", "MenuItem"));
```
FedeIlLeone, line 96: Use waitForProps and the type parameter

src/renderer/coremods/contextMenu/index.tsx
```diff
@@ -88,53 +91,69 @@ type ContextMenuData = ContextMenuProps["ContextMenu"] & {
  * @internal
```
FedeIlLeone, line 91: You can remove this comment now

src/renderer/coremods/contextMenu/index.tsx
```diff
@@ -88,53 +91,69 @@ type ContextMenuData = ContextMenuProps["ContextMenu"] & {
...
+        }
+      });
+
+      return <ContextMenu {...menu} />;
+    },
+  );
+}
+export function start(): void {
+  injector = new Injector();
+  void injectMenu();
```
FedeIlLeone, line 154: You can move the implementation in start, a dedicated function is not needed in this case.

FedeIlLeone (changes requested):

Tharki-God:
moved back to plain text patch to get less rerenders.

src/renderer/coremods/contextMenu/index.tsx
```diff
@@ -84,57 +84,74 @@ type ContextMenuData = ContextMenuProps["ContextMenu"] & {
...
+      const itemRet = makeItem(item.getItem(data, menu)) as ContextItem & {
+        props: { replug?: boolean };
+      };
+
+      // removed the code to add unique id every time because it made experience bad while hovering over item and it getting updated
+      // should be fine because we actively keep removing old items ig
 
-      if (res?.props) {
-        // add in unique ids
-        res.props.id = `${res.props.id || "repluggedItem"}-${Math.random()
```
FedeIlLeone, line 128: Can you add the unique id back. In what way "it made experience bad while hovering over item and it getting updated". I'm not having any "bad experience" with it.

src/renderer/coremods/contextMenu/index.tsx
```diff
@@ -84,57 +84,74 @@ type ContextMenuData = ContextMenuProps["ContextMenu"] & {
...
-          .toString(36)
-          .substring(2)}`;
+      if (itemRet?.props) {
+        // add in prop  for cleanup
+        itemRet.props.replug = true;
```
FedeIlLeone, line 147: `replug` prop is not needed

src/renderer/coremods/contextMenu/plaintextPatches.ts
```diff
@@ -7,7 +7,10 @@ export default [
       {
         match: /var \w,\w=(.)\.navId/,
         replace: (vars, menu) =>
-          `replugged.coremods.coremods.contextMenu._insertMenuItems(${menu});${vars}`,
+          `if (!${menu}.plugged){` +
```
FedeIlLeone, line 10: Do these checks in the function itself.

FedeIlLeone (changes requested):

src/renderer/coremods/contextMenu/index.tsx
Tharki-God: This is how ig. you need an item that gets updated and if you add unique ID everytime then it will be different from what hover had in its state and this will happen

src/renderer/coremods/contextMenu/index.tsx
FedeIlLeone: It is never checking if this prop exists or not

src/renderer/coremods/contextMenu/plaintextPatches.ts
Tharki-God: okay first check if statement is to check if it's remade menu which can't be done inside function or else it will be infinite loop kinda. second if statement is there to check if we returned null/undefined and that's only if we can't get the this function from module itself so this can't be inside function either.
if you know how this can be inside function then I am open, but to my knowledge it needs to be handled here.

src/renderer/coremods/contextMenu/index.tsx
Tharki-God: check on the filter above this

src/renderer/coremods/contextMenu/index.tsx
FedeIlLeone: I have tested and is not changing id.

src/renderer/coremods/contextMenu/index.tsx
FedeIlLeone: No. That filter is checking for the `plugged` property, not `replug`.

Tharki-God: yeah I messed up and fixed now ig

src/renderer/coremods/contextMenu/index.tsx
Tharki-God: > I have tested and is not changing id.

you updating the menu?

FedeIlLeone: Unfortunately we need to have the unique id, there might be some plugins that use the same id...

src/renderer/coremods/contextMenu/index.tsx
FedeIlLeone: This is causing issues with submenus.

src/renderer/coremods/contextMenu/index.tsx
Tharki-God: > Unfortunately we need to have the unique id, there might be some plugins that use the same id...

plugin devs should address that not the client mod, we should value user experience more.
idk will have to think of some way to store it or anything then to keep old id's maybe a map on the menu props?

FedeIlLeone: This will result in a breaking change if you remove the unique id, there are already some plugins that are adding items without an id because we were handling that with the API. Yes, we need to find a better solution to save the id, take a look at how it has been done with message popover API.

Tharki-God: even there it's not a unique ID always. either they provide a id or random ID. we can do that here too I mean but unique always is just meh kinda but we can store them on the menu as it stays constant just the children gets changed and the plugged value on it

src/renderer/coremods/contextMenu/index.tsx
```diff
@@ -16,26 +16,36 @@ export const menuItems = {} as Record<
...
   }
   if (React.isValidElement(raw)) {
     // We can't construct something that's already made
-    return raw as ContextItem | undefined;
+    return raw as RepluggedContextItem;
   }
 
   const { type, ...props } = raw;
+  // add in prop  for cleanup
+  props.replug = true;
```
FedeIlLeone, line 46: You may want to move this back in `_buildPatchedMenu`, updating menu checkboxes doesn't delete them.

FedeIlLeone (changes requested):

src/renderer/coremods/contextMenu/index.tsx
```diff
@@ -78,7 +85,7 @@ export function removeContextMenuItem(navId: ContextMenuTypes, getItem: GetConte
 }
 
 type ContextMenuData = ContextMenuProps["ContextMenu"] & {
-  children: React.ReactElement | React.ReactElement[];
+  children: React.ReactElement[];
```
asportnoy, line 88: Why?

Tharki-God: because I make it an array if it's just a single element anyway.

asportnoy: But then users can't input a single element

Tharki-God: - they can? wdym, this is type of what we get from api not anything users input

src/renderer/coremods/contextMenu/index.tsx
```diff
@@ -88,68 +95,134 @@ type ContextMenuData = ContextMenuProps["ContextMenu"] & {
...
-    ?.props?.children?.props?.id?.startsWith("devmode-copy-id-");
-  if (hasCopyId) {
-    menu.children.splice(-1, 0, repluggedGroup);
+  // make children array if it's not already
+  if (!Array.isArray(menu.children)) {
+    menu.children = [menu.children];
+    if (!Array.isArray(menu.children)) {
+      return <ContextMenu {...menu} plugged={true} />;
+    }
+  }
```
12944qwerty, line 126: Why are you checking if it's an array after making it an array?

src/renderer/coremods/contextMenu/index.tsx
```diff
@@ -88,68 +95,134 @@ type ContextMenuData = ContextMenuProps["ContextMenu"] & {
...
+
+    // Add in the new menu items right above the DevMode Copy ID
+    // If the user doesn't have DevMode enabled, the new items will be at the bottom
+    const hasCopyId =
+      menu.children.at(-1)?.props?.children?.props?.id?.startsWith("devmode-copy-id-") ||
+      menu.children
+        .at(-1)
+        ?.props?.children?.some(
+          (c: React.ReactElement | null) => c?.props?.id?.startsWith("devmode-copy-id-"),
+        );
```
12944qwerty, line 141: Maybe use `getByProps("DeveloperMode").DeveloperMode.getSetting()`?

src/renderer/coremods/contextMenu/plaintextPatches.ts
```diff
@@ -7,7 +7,11 @@ export default [
       {
         match: /((\w+)\){)(var\s*\w+;let{navId:)/,
         replace: (_, prefix, menu, suffix) =>
-          `${prefix}replugged.coremods.coremods.contextMenu._insertMenuItems(${menu});${suffix}`,
+          `${prefix}const patchedMenu=replugged.coremods.coremods.contextMenu._buildPatchedMenu(${menu});${suffix}`,
+      },
+      {
+        match: /(patchedMenu.{2400,2600}return)(\(0,.\.jsx\)\(\w+.OnMenuSelectContext)/,
```
12944qwerty, line 13: I can't find this match in sources? Use a find if possible too

12944qwerty (changes requested):

src/renderer/coremods/contextMenu/plaintextPatches.ts
TharkiDev: oh thats because a match works only if the first one replaced. 
i am using find btw?

src/renderer/coremods/contextMenu/index.tsx
TharkiDev: not all content menus would have that yk

src/renderer/coremods/contextMenu/index.tsx
TharkiDev: i dont remember. maybe i will remove it then

src/renderer/coremods/contextMenu/plaintextPatches.ts
12944qwerty: Why would you do it like that? Why not replace it all at once correctly...?

TharkiDev: because first one runs in start of function and second in the end where it returns

YofukashiNo: solved

yofukashino:
closing in favor of https://github.com/replugged-org/replugged/pull/651

Issue #350, opened 2023-01-21
Splash screen theming


asportnoy:

bop34:
what does this involve?

PR #523, opened 2023-07-22
Implement splash themes

EastArctica:
This PR implements a preloader for the splash screen, currently it only loads themes as there isn't much else to load on there or any use for it.

Things of note:
 - Should the following be required from the `renerer` instead of from `splashrenderer`? `util.ts`, `modules/logger.ts`, `managers/quick-css.ts`, `managers/themes.ts` and `apis/settings.ts`. Personally I don't think so, In my opinion the splashrenderer and main renderer should be separate.
 - Feel free to suggest a better or different name for splashrenderer.
 
Closes: #350


cspell.json
```diff
@@ -61,7 +61,8 @@
     "Vendicated",
     "Webauthn",
     "weblate",
-    "XLARGE"
+    "XLARGE",
+    "splashrenderer"
```
asportnoy, line 65: Use camelCase for the directory name, then you won't need this.

src/preloadSplash.ts
asportnoy, line 1: There is a LOT of duplicated code in this file and the ones below. We should not have to maintain two copies of the same code. Please figure out some way to dedupe it.

asportnoy (changes requested):

src/preloadSplash.ts
EastArctica: I can remove the RepluggedNative and instead only expose the ignitor if that would be better(it would)? But I think that preloadSplash should be separate from preload even though it would reduce the duplicated code. As for the files mentioned in the description of the PR, what would you like to do about those?

cspell.json
EastArctica: 👍 Will do, thanks for the info!

src/preloadSplash.ts
asportnoy: I don't see a reason why we shouldn't just import them from the splash screen

EastArctica: You mean import the files from preload/... in preloadSplash? If so I can implement that when I get home(5 pm est). If I remember correctly I think one of them differs from the one in preload but I think that one is fine to be in preloadSplash?

asportnoy: Yes. If there are any differences, those can be changed. But you might also be able to tweak the original file.

EastArctica:
Alright well, this is how it should've been implemented in the beginning in my opinion (as long as you're okay with what's going on in src/renderer/index.ts, I don't love it)

src/renderer/index.ts
```diff
@@ -2,5 +2,10 @@ import * as replugged from "./replugged";
 
 window.replugged = replugged;
 
-await replugged.plugins.loadAll();
-await replugged.ignition.ignite();
+// Splash screen
+if (document.title === "Discord Updater") {
```
asportnoy, line 6: I don't think this will work with non-english languages. Is there a better way you can determine this?

src/renderer/managers/ignition.ts
```diff
@@ -75,3 +75,19 @@ export async function ignite(): Promise<void> {
   await componentsReady();
   await start();
 }
+
+export async function startSplash(): Promise<void> {
+  log("Ignition", "Start", void 0, "Igniting Replugged Splash Screen...");
+  const startTime = performance.now();
+
+  await themes.loadMissing().then(themes.loadAll);
```
asportnoy, line 83: This will load the main.css file, not the spash.css file.

src/renderer/managers/ignition.ts
```diff
@@ -75,3 +75,19 @@ export async function ignite(): Promise<void> {
...
+  // Quick CSS needs to be called after themes are loaded so that it will override the theme's CSS
+  quickCSS.load();
```
asportnoy, line 85: I don't think we want quick css here

asportnoy (changes requested):

src/renderer/managers/ignition.ts
EastArctica: You mean no quickcss should be executed in the splash screen?

src/renderer/index.ts
EastArctica: Will do, good catch

EastArctica: Ended up not being needed, although I found a better way anyway.

EastArctica:
Tested on windows, can't see any reason it would fail on any other platform 🤷

lexisother (approved):

lexisother:
Superseded by #560.

PR #530, opened 2023-08-07
Addon version release command

asportnoy:
Adds a command to easily release a new version of an addon. Will prompt you for a version number (or semver update type), update the manifest and package, commit and tag the changes, and push to the remote, with various prompts along the way. 

I did consider using [isomorphic-git](https://isomorphic-git.org/) at first, but it can't use the local git config for authentication, commit author, etc, which makes it impractical to use, so I decided executing the git commands directly was the best bet.


bin/release.mjs
```diff
@@ -0,0 +1,259 @@
...
+      },
+    },
+    { onCancel },
+  );
+
+  // Commit changes
+  runCommand(`git commit -m "${message}"`);
+
+  // Tag commit
+  runCommand(`git tag ${tagName}`);
```
lexisother, line 250: Allow signing the tag. I know this can be set as a global git config, but I'd prefer to have it be passable as well. The flag is `-s`.

lexisother (changes requested):

lexisother:
Is this ready?

asportnoy:
@lexisother once you approve, yep

lexisother (approved):

PR #531, opened 2023-08-07
Remove line breaks in plaintext patches

Tharki-God:
removes all line breaks from code in plain text patch. wouldnt break anything since minified codes work fine without line breaks


lexisother (dismissed):

asportnoy (approved):

asportnoy (changes requested):
This needs to be updated to handle literal `\n`s in the code

colin273 (changes requested):
Why would we modify Discord's code beyond what we already do? This makes assumptions about Discord's minifier that I would prefer to avoid, namely that newline characters will never appear in, say, a template string. Perhaps this is true, but if it ever stops being true, I don't think this is something we should have to go back and fix.

I would prefer that developers account for possible white space in their regexes as JavaScript syntax would logically allow, if they're willing to attempt plaintext patches at all.

lexisother:
If you as a plugin developer come across an instance where a newline in code is messing with your plaintext patch (e.g. when using `.` to target characters), move to `[^]` and manually account for other instances of whitespace.

PR #534, opened 2023-08-14
Better implementation of useStateFromStoresArray and useStateFromStoresObject

FedeIlLeone:
`useStateFromStoresArray` recently became undefined and together with `useStateFromStoresObject` both had an unstable regex. ~~Recreating the function used by the original hooks should be better.~~ Uses Lodash's `isEqual` function.

Includes updated types for `ChannelMessages` and new navIds for `ContextMenuTypes` enum.

### DIFF
--- a/src/renderer/modules/common/flux.ts
+++ b/src/renderer/modules/common/flux.ts
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import type { DispatchBand, FluxDispatcher as Dispatcher } from "./fluxDispatcher";
 import { filters, getFunctionBySource, waitForModule, waitForProps } from "../webpack";
+import type { DispatchBand, FluxDispatcher as Dispatcher } from "./fluxDispatcher";
 
 type DispatchToken = string;
 type ActionType = string;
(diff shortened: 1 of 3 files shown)


src/renderer/modules/common/flux.ts
```diff
@@ -183,34 +183,71 @@ const SnapshotStoreClass = await waitForModule<typeof SnapshotStore>(
...
+  const keysB = Object.keys(b);
+
+  if (keysA.length !== keysB.length) return false;
+
+  for (let i = 0; i < keysA.length; i++) {
+    let key = keysA[i];
+    if (a[key] !== b[key]) return false;
+  }
+
+  return true;
```
asportnoy, line 199: You can use lodash `xor` with the keys here - https://stackoverflow.com/a/54788640/6911703

FedeIlLeone:
Well actually now they both do the same thing. I might just remove them and if a dev needs uSFSArray/uSFSObject they can just pass `_.isEqual` to the `useStateFromStores` comparator. 🤷🏻‍♂️

asportnoy (approved):

PR #537, opened 2023-08-21
Update fluxDispatcher.ts

Catalyst4222:
`fluxDispatcher._actionHandlers.getOrderedActionHandlers` returns an array of Handlers, not `void`

image
image


FedeIlLeone (approved):
Most likely I missed this one when I was typing it.

PR #540, opened 2023-08-22
Add transparency option

EastArctica:
This PR adds a transparency button into settings. This has been confirmed tested on Manjaro linux as well as I believe tested on Windows although that should be double checked.

Things of note:
 - ~~In `src/main/index.ts`, we manually get the replugged settings. This is needed in order for the launcher to know whether to launch supporting transparency or not. This should be moved elsewhere or done differently however I would like some input on it.~~
 - This does not apply to the splash screen. If/When #523 is implemented, that should be added.
 - ~~Currently there is not translation key for `REPLUGGED_SETTINGS_TRANSPARENT` or `REPLUGGED_SETTINGS_TRANSPARENT_DESC`. After speaking with @asportnoy, we determined that the various issues with having transparency should be noted per platform instead of all together as they were pre-swc.~~
 - ~~The settings is currently under advanced settings, however I suppose there is no reason for this to be and it should be moved out and into the general settings.~~
 
 If anyone encounters any bugs with this please let me know!


src/main/index.ts
asportnoy, line 1: There is a `getSetting` function already in `src/main/ipc/settings.ts` that you can use

src/renderer/coremods/settings/pages/General.tsx
```diff
@@ -173,6 +177,16 @@ export const General = (): React.ReactElement => {
           }}>
           {Messages.REPLUGGED_SETTINGS_DEV_COMPANION}
         </ButtonItem>
+
```
asportnoy, line 180: This shouldn't be under advanced IMO

asportnoy (changes requested):

EastArctica:
Opinions on these translation keys? The linux/windows one would be used depending on which platform you're on.
```
REPLUGGED_SETTINGS_TRANSPARENT: `Transparent Window`
REPLUGGED_SETTINGS_TRANSPARENT_DESC: `Makes the discord window transparent, primarily useful for theming.`
REPLUGGED_SETTINGS_TRANSPARENT_ISSUES_WINDOWS: `****WARNING:**** This will break **window snapping**. In some cases, you may experience a black background, such as when the window is cut off at the top or bottom due of the monitor resolution, or when the development tools are open and docked. **Requires restart**.`
REPLUGGED_SETTINGS_TRANSPARENT_ISSUES_LINUX: `****WARNING:**** **Hardware acceleration** may need to be turned **off**. In some cases, you may experience a black background, such as when the window is cut off at the top or bottom due of the monitor resolution, or when the development tools are open and docked. **Requires restart**.`
```

EastArctica:
Should be all set, needs testing on windows and macos and then a review.

EastArctica:
Confirmed working on windows after minor issues, just need test on macos.

EastArctica:
Windows support seems to work. MacOS has been tested once and worked(via manually triggering `RepluggedNative.transparency.setVibrancy("under-window")`) but not on the latest commit which fixes an issue with the styles being read incorrectly.

## Splash Screen
Currently, the splash screen is neither blurred nor transparent. This could provide an inconsistently between splash theming and normal theming. However, implementing blur into the splash screen could be difficult as core mods don't load there and therefore, we could not detect when the css is loaded to apply the blur.

## Blur options
(Theoretically) supported vibrancy options(macOS):
`titlebar`, `selection`, `menu`, `popover`, `sidebar`, `header`, `sheet`, `window`, `hud`, `fullscreen-ui`, `tooltip`, `content`, `under-window`, and `under-page`

The following are deprecated and have been removed in macOS Cataline (10.15):
`appearance-based`, `light`, `dark`, `medium-light`, and `ultra-dark`

Supported effect options(Windows):
`blurbehind`, `acrylic`, `mica`, and `unified-acrylic`

## Windows
Windows support is iffy and seems to be a "whatever works for you" situation.
Currently, the warning message on windows is as follows:
> ****WARNING:**** This will break **window snapping**. In some cases, you may experience a black background, such as when the window is cut off at the top or bottom due to the monitor resolution, or when the development tools are open and docked.

We may want to change this to mention that some blur options may not be supported on one's system.

Due to these issues, below is a list of the tested versions.
### Tested Versions
Windows 10 22H2:
`unified-acrylic` - Lags behind cursor, Blurs correctly
`blurbehind` - Works perfectly
`acylic` - Lags behind cursor, Blurs correctly
`mica` - No background, not transparent.

## General setup
For blur to work, both `html` and `body` must have `background: transparent`. Below is an example of setting up blur on both macOS and Windows.
```css
html, body {
  background: transparent !important;
}

:root {
    --window-vibrancy: "under-window";
    --window-win-blur: "blurbehind";
}
```

## Linux
Currently the Linux support is... iffy... I have no plans to add Linux support as the majority of the userbase is on macOS or Windows. Additionally, I have added the basics to allow the user to add the blur themselves. Enabling the transparency effect in settings will make the window background 100% transparent (with the same requirements as above). However, it will have no blue. Instead, I'm leaving this up to the user to manage with their compositor of choice.

The current Linux specific warning message is this:
> ****WARNING:**** **Hardware acceleration** may need to be turned **off**. In some cases, you may experience a black background, such as when the window is cut off at the top or bottom due to the monitor resolution, or when the development tools are open and docked.

Perhaps we should change this to mention that blur is not supported and needs to be implemented by the user?

SiriusBYT:
## Tested OS: Windows 11 (Enterprise N) Canary Channel Build 26052 (24H2):
### Discord Canary:
- Unstable, transparency effect does not work in the slightest
- Causes "Bloom", the background is black and every element has a blur behind them, changing the blur effects doesn't change anything

### Discord PTB:
- Crashes on startup
DiscordPTB_bevZ0kRKba
### Discord Stable:
- Only "unified-acrylic" works
- "blurbehind" only works when dragging the window but the background doesn't gets blurred at all.
- Acrylic works indeed but does in fact, lag like hell when dragging.
- Opening pins causes the transparency to shit itself, don't know why.
Discord_xtbB7Z0h1U
- The following code is not enough to trigger transparency:
```
html, body {
  background: transparent !important;
}
```
Too lazy to figure out which variables make it fully work exactly, will see later when I implement this into FC, but this code seems to do the trick even on Nitro themes:
```
html, body {
    --bg-overlay-color: transparent;
    --background-primary: transparent;
    --background-secondary: transparent;
    --background-secondary-alt: transparent;
    --background-tertiary: transparent;
    --bg-overlay-1: transparent;
    --bg-overlay-2: transparent;
    --bg-overlay-3: transparent;
    --bg-overlay-4: transparent;
    --bg-overlay-5: transparent;
    --bg-overlay-6: transparent;
    --bg-overlay-hover: transparent;
    --bg-overlay-active: transparent;
    --bg-overlay-selected: transparent;
    --bg-overlay-chat: transparent;
    --bg-overlay-home: transparent;
    --bg-overlay-home-card: transparent;
    --bg-overlay-app-frame: transparent;

    background: transparent !important;
}
```
- Transparency has this weird "snake" bug, don't know how to explain it other than this video (Opacity issue?).

https://github.com/replugged-org/replugged/assets/87911136/70f3ed0d-153d-45dc-bb8a-a0dcfedf3f3c

### Other notes
- The advanced settings under the general tab of Replugged puts you into a crash loop
- Transparency takes time to start, causing the client to be black for a few seconds

EastArctica:
Fixed some issues, looks like blur isn't possible with neither the native [setBackgroundMaterial](https://www.electronjs.org/docs/latest/api/browser-window#winsetbackgroundmaterialmaterial-windows) nor [@pyke/vibe](https://github.com/pykeio/vibe)

barutu (approved):

zrodevkaan (commented):
nuh uh

EastArctica:
If anyone would like to take this over please just let me know.

EastArctica:
Example video of it working: https://zipline.eastarcti.ca/u/KgFiYH.mp4
Minimum needed for the transparency effect to trigger (Please note this does not mean the minimum needed for you to *see* transparency):
```css
:root {
  --window-background-material: "acrylic";
  --window-background-color: #00000000;
}

html, body {
  background: transparent;
}
```

Example theme using acrylic https://github.com/EastArctica/replugged-nord-transparent/releases/tag/v1.0.3

Ren0X1:
maybe we  can do something listening the resize event from the BrowserWindow object something like this:
`window.on('resize', () => {
  window.webContents.send('resize-event');
});`

EastArctica:
@Ren0X1 This probably isn't what you meant but while looking into it the latest commit changes the way maximize/unmaximize is done in a few ways but seems to prevent the window from being at pixel 1,1 while still fixing the grey background as well as making the maximize experience a little bit cleaner.

If anyone knows/can figure out how to unmaximize when dragged while in the maximized state cleanly please let me know!

Ren0X1 (approved):

PR #541, opened 2023-08-27
Update ESBuild and convert bin to TypeScript

asportnoy:
- Update ESBuild to v0.19 (latest). This is needed for some new TS features we want to use. There were some minor breaking changes here that should now be handled.
- Convert bin to TypeScript via ESBuild. Adds a `postinstall` script to compile it.
- Remove Parcel dependency, switch to using ESBuild for compiling CSS/SCSS. An advantage of this is we will not need to add parcel as a dependency inside of themes.


lexisother (approved):
Sure.

PR #543, opened 2023-08-29
Call injected callbacks with the this keyword set

Penguin-Spy:
As the title says, calls injected functions with the `this` keyword set to what the original function was called with. The `self` parameter is left for backwards compatibility.


asportnoy (commented):
See https://discord.com/channels/1000926524452647132/1000955965304221728/1146612714186211368

asportnoy (approved):

PR #546, opened 2023-09-07
type error fix

Tharki-God:
fixes an type error while using use settings with radio item


asportnoy (changes requested):
If you have to use types and not interfaces, you need to disable the ESLint rule for the appropriate lines

Tharki-God:
> If you have to use types and not interfaces, you need to disable the ESLint rule for the appropriate lines

doesnt show any error to me, why do we need to disable eslint then?

lexisother:
> doesnt show any error to me, why do we need to disable eslint then?

https://github.com/replugged-org/replugged/actions/runs/6115469346/job/16599047576#step:8:12

asportnoy (approved):

src/renderer/modules/components/RadioItem.tsx
```diff
@@ -13,9 +14,10 @@ interface RadioOptionType {
   tooltipPosition?: "top" | "bottom" | "left" | "right" | "center" | "window_center";
   icon?: React.ComponentType<unknown>;
   collapsibleContent?: React.ReactNode;
-}
+};
 
-interface RadioProps {
+// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
+type RadioProps = {
```
FedeIlLeone, line 20: Changing this is not required.

FedeIlLeone (changes requested):

Issue #528, opened 2023-08-04
Install addon command


asportnoy:
[blocked until the commands API is finished]

We should add a command to open the install flow: `/install <id> [source]`. Then we can share this command in the store pages as an alternative when a user's browser is unable to connect.

asportnoy:
Yours now

trollo2:
anything new

asportnoy:
Haven't had a chance to rest after the latest update but this is still planned for the v4.7.0 update

PR #551, opened 2023-09-15
Install Command

ShadiestGoat:
Hello!

This is a WIP PR for the install commands (#528). At the time of creation, there is an install command that works, but I'm honestly not sure what to do with translations. I marked at least 11 strings that may need translations, and I couldn't really find anything that already exists.


ShadiestGoat:
Oh also - should've mentioned: in the original issue, the proposed syntax was `/install <id> [source]`, however, this doesn't account for monorepos, and may introduce further confusion with regards to identifier vs id, so I have `/install <identifier> [source] [id]`.

In the same spirit, I'm not sure if the id was meant to be the `updater.id` or the manifest id, but the updater id is used to find the repo so I assumed its that one

Tharki-God:
add a share command too if you doing the pr on this ig

src/renderer/coremods/installer/commands.ts
```diff
@@ -0,0 +1,81 @@
+import { Injector } from "@replugged";
+import { ApplicationCommandOptionType } from "src/types";
+import { INSTALLER_SOURCES, InstallerSource, getInfo, install } from "./util";
+
+const injector = new Injector();
```
asportnoy, line 5: Don't create a new injector here, just export the one from index.tsx and re-use it. Then you don't need to deal with uninjecting separately.

src/renderer/coremods/installer/commands.ts
```diff
@@ -0,0 +1,81 @@
...
+
+const installSourceDisplayName: Record<InstallerSource, string> = {
+  github: "GitHub",
+  store: "Store",
+};
```
asportnoy, line 10: I think these will also need i18n.

src/renderer/coremods/installer/commands.ts
```diff
@@ -0,0 +1,81 @@
...
+
+      if (!info) {
+        return {
+          send: false,
+          // TODO: i18n?
+          result: "Error: couldn't fetch the plugin",
+        };
+      }
+
+      const success = await install(info);
```
asportnoy, line 68: Make the command show the install prompt like the other install methods.

asportnoy (changes requested):
I will deal with adding the i18n strings and let you know so you can swap them out. Some additional feedback below for now.

src/renderer/coremods/installer/commands.ts
ShadiestGoat: Unless I extract the injector from index into a 'common' file, there would be dependency cycles or miss placed variables, which would be kind of a mess.

I think keeping a new injector here is the cleanest option?

asportnoy: You could pass the injector as an arg to the function

ShadiestGoat: True, let me get on that

src/renderer/coremods/installer/commands.ts
```diff
@@ -0,0 +1,116 @@
...
+
+function installSourceName(source: InstallerSource): string {
+  const displayName = specialSourceDisplayNames[source];
+  if (displayName) {
+    return displayName;
+  }
+
+  // TODO: i18n
+  return source;
+}
```
asportnoy, line 20: I think this is way overcomplicated. Just have a display name map like you had before and hardcode the value for GH.

src/renderer/coremods/installer/commands.ts
```diff
@@ -0,0 +1,116 @@
...
+      },
+    ],
+
+    async execute(args) {
+      await installFlow(
+        // @ts-expect-error identifier is a required argument, so no need for assertion
+        // eslint-disable-next-line @typescript-eslint/non-nullable-type-assertion-style
+        args.find((v) => v.name === "identifier").value as string,
+        args.find((v) => v.name === "source")?.value as InstallerSource | undefined,
+        args.find((v) => v.name === "id")?.value,
```
asportnoy, line 68: Use getValue like we have in the commands file. It will take care of TS for you.

src/renderer/coremods/installer/commands.ts
ShadiestGoat: The problem with that is translation. If theres a display name map, we can't use translations, right?

The other alternative is to remove the map in the first place, and only keep the installSourceName function. This way would be able to 'map' translation values, hard code some values like GitHub, and be easy to expand on

asportnoy: Why wouldn't we be able to use translations? Anywhere we use a string to display text we should be able to display an i18n value instead.

src/renderer/coremods/installer/commands.ts
ShadiestGoat: args doesn't have getValue. This is in an `execute` function, not an `executor` (with execute you can just not give a response, I don't think thats possible with `executor`?)

asportnoy: I'm pretty sure you can just return undefined?

ShadiestGoat: Not according to the types. Although, nullish values *are* handled. Should I include a void return type in this PR?

Tharki-God: just return an object with send value false I and nothing else would say

src/renderer/coremods/installer/commands.ts
ShadiestGoat: When are the translations initialized? If use this:

```ts
const sourceDisplayNames = {
   "store": Messages.*
}
```

Then wouldn't it be initialized with the wrong string?

In that case we could load it at start() but that might not be best practice?

asportnoy: I think it wouldn't be an issue but you can test and see.

ShadiestGoat: Oh wow, you are actually correct. Thats odd I really didn't expect it, I'll update in a sec

ShadiestGoat:
Wait, I just realized the share cmd uses plugins manager so themes will not work

asportnoy (changes requested):
- Implement i18n strings, now added on upstream
- Rename identifier to `addon` to match i18n
- Allow install links to be specified for addon. If the other parameters aren't specified, that will   be used as a fallback, while also being used to determine the addon identifier.
- Disable toasts for the install flow errors, and instead use the command responses for that.

ShadiestGoat:
For point 3, you should probobly update the i18n string to say that urls are also allowed

src/renderer/coremods/installer/commands.ts
```diff
@@ -0,0 +1,88 @@
...
+      try {
+        // If addon is not a url, this will throw an error
+        const _ = new URL(addon);
+        const resp = parseInstallLink(addon);
+        if (resp) {
+          addon = resp.identifier;
+          source = resp.source;
+          id = resp.id;
+        }
+      } catch {}
```
asportnoy, line 62: The URL check here is unnecessary. `parseInstallLink` should already handle that and will return null. Remove the try/catch and the unnecessary URL constructor.

asportnoy (approved):

Issue #355, opened 2023-01-23
Include error information in ErrorBoundary message


asportnoy:
We should try to include the message and the stack trace and maybe some other relevant info if we have it.

PR #553, opened 2023-09-16
feat(ErrorBoundary): add error stack and plugin name

FedeIlLeone:
- Extract and show the plugin name if the stack includes it.
- Show the error stack trace in a collapsible section.

Closes: #355


i18n/en-US.json
```diff
@@ -232,5 +232,6 @@
...
     "REPLUGGED_SETTINGS_TRANSPARENT_DESC": "Makes the Discord window transparent, primarily useful for theming. **Requires restart**.",
     "REPLUGGED_SETTINGS_TRANSPARENT_ISSUES_WINDOWS": "****WARNING:**** This will break **window snapping**. In some cases, you may experience a black background, such as when the window is cut off at the top or bottom due to the monitor resolution, or when the development tools are open and docked.",
-    "REPLUGGED_SETTINGS_TRANSPARENT_ISSUES_LINUX": "****WARNING:**** **Hardware acceleration** may need to be turned **off**. In some cases, you may experience a black background, such as when the window is cut off at the top or bottom due to the monitor resolution, or when the development tools are open and docked."
+    "REPLUGGED_SETTINGS_TRANSPARENT_ISSUES_LINUX": "****WARNING:**** **Hardware acceleration** may need to be turned **off**. In some cases, you may experience a black background, such as when the window is cut off at the top or bottom due to the monitor resolution, or when the development tools are open and docked.",
+    "REPLUGGED_SETTINGS_ERROR_PLUGIN_NAME": "Plugin: {name}"
```
asportnoy, line 236: Add in Weblate, not here

src/renderer/modules/components/ErrorBoundary.tsx
```diff
@@ -40,11 +86,19 @@ export default class ErrorBoundary extends React.Component<ErrorProps, ErrorStat
...
     if (this.state.hasError) {
+      const { error, pluginName } = this.state;
+
       return (
         this.props.fallback || (
           <div className="replugged-error-boundary">
             <h1>{Messages.REPLUGGED_SETTINGS_ERROR_HEADER}</h1>
+            {pluginName && (
+              <p className="replugged-error-boundary-plugin">
+                {Messages.REPLUGGED_SETTINGS_ERROR_PLUGIN_NAME.format({ name: pluginName })}
```
asportnoy, line 97: Just to make sure this won't error
```suggestion
                {Messages.REPLUGGED_SETTINGS_ERROR_PLUGIN_NAME?.format?.({ name: pluginName })}
```

src/renderer/modules/components/ErrorBoundary.tsx
```diff
@@ -40,11 +86,19 @@ export default class ErrorBoundary extends React.Component<ErrorProps, ErrorStat
...
+              </p>
+            )}
             <p>{Messages.REPLUGGED_SETTINGS_ERROR_SUB_HEADER}</p>
+            {error?.stack && <CollapsibleErrorStack stack={error.stack} />}
```
asportnoy, line 101: Wrap this in an errorboundary with empty fallback in case it somehow errors itself

i18n/en-US.json
FedeIlLeone: I did.

asportnoy (approved):

Tharki-God:
okay just an request but can you make it so if its an react error which it probably would be then show the actual error instead of code and link?

PR #554, opened 2023-09-17
Proper imports from types package

### LINKED ISSUE #247: [Plugin template] Fix importing a path
Let's say I want to import `replugged.webpack.filters`. VSCode will automatically add:
```js
import { filters } from "replugged/dist/renderer/modules/webpack";
```
However, we do not currently support paths on the export. So we'd have to do something like:
```js
import { webpack } from "replugged";
const { filters } = webpack;
```

Ideally, we should support the first method. However, if we can't figure it out for some reason. we should figure out how to prevent VSCode from suggesting it.

colin273:
Allows plugin devs to import things from replugged's types package and have those imports transformed to references to window.replugged seamlessly. No more importing and then destructuring. Closes #247.

### DIFF
--- a/bin/replugged.mjs
+++ b/bin/replugged.mjs
@@ -257,20 +257,7 @@ async function buildPlugin({ watch, noInstall, production, noReload }) {
     // @ts-expect-error
     setup: (build) => {
       // @ts-expect-error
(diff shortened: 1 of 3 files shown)


colin273:
making a new PR to avoid merge conflicts

PR #555, opened 2023-09-17
Support sub-imports from replugged package

### LINKED ISSUE #247: [Plugin template] Fix importing a path

colin273:
Finally, no more painful destructuring. Closes #247.


scripts/create-import-wrappers.mts
```diff
@@ -0,0 +1,15 @@
...
+const locations = {
+  modules: ["logger", "webpack", "i18n", "injector", "common", "components"],
+  apis: ["commands", "notices", "settings"],
+  util: ["util"],
+};
+
+for (const [category, names] of Object.entries(locations)) {
+  for (const name of names) {
+    const path = `./dist/renderer/${category}/${name}`;
+    const dtsContents = `import * as exp from "${path}"; export = exp;`;
```
asportnoy, line 12:
```suggestion
    const dtsContents = `export * from "${path}";`;
```

asportnoy (changes requested):
- Add `/*.d.ts` to the gitignore
- VSCode still suggests importing importing from `replugged/dist`, so the build script should still handle that. (It still needs to be available for importing types I believe so we can't just completely exclude it.

Tharki-God:
> * VSCode still suggests importing importing from `replugged/dist`, so the build script should still handle that. (It still needs to be available for importing types I believe so we can't just completely exclude it.

I think path aliases would do the trick here.

colin273:
> [`replugged/dist`] still needs to be available for importing types I believe so we can't just completely exclude it.
I'll make another import wrapper for types and then make the build script reject imports from dist. All our types are exported from the top level of the types directory.

Tharki-God:
> > [`replugged/dist`] still needs to be available for importing types I believe so we can't just completely exclude it.
> 
> I'll make another import wrapper for types and then make the build script reject imports from dist. All our types are exported from the top level of the types directory.

what about the types for example store? they aint exported from top level of types?

colin273:
> what about the types for example store? they aint exported from top level of types?

I'm not sure which type you're referring to, but everything that is defined in src/types is exported from its top level.

Tharki-God:
> > what about the types for example store? they aint exported from top level of types?
> 
> I'm not sure which type you're referring to, but everything that is defined in src/types is exported from its top level.

this for example `import type { Store } from "replugged/dist/renderer/modules/common/flux";`

colin273:
Those imports will still work if you want to use them, but you can also import flux from replugged/common. If it's not part of the types directory, it does not matter whatsoever for this particular re-export.

asportnoy:
I tried importing it from common earlier in the way you suggested and it didn't seem to work. Are you sure that works?

colin273:
`import { flux } from "replugged/common"` works for me. Yes I'm sure. Importing from `replugged/common/flux` does not, obviously.

Tharki-God:
> `import { flux } from "replugged/common"` works for me. Yes I'm sure. Importing from `replugged/common/flux` does not, obviously.

you cant take type from it tho

colin273:
Who cares. Keep importing from dist or just use `flux.Store`.

Tharki-God:
> Who cares. Keep importing from dist or just use `flux.Store`.

wait flux.Store would work? idts but do verify and tell

Tharki-God:
i think these kind of types are only on common so can't we redirect the `replugged/dist/renderer/modules/common/*` to `replugged/common/types/*` with path aliases or anything?

colin273:
Path aliases as in the tsconfig things? Those don't work when this is an npm package installed in some other repo's node_modules. I tried that, no luck. Would be nice, but absolutely not happening.

Tharki-God:
> Path aliases as in the tsconfig things? Those don't work when this is an npm package installed in some other repo's node_modules. I tried that, no luck. Would be nice, but absolutely not happening.

i think type imports are stripped from code at build

colin273:
> wait flux.Store would work? idts but do verify and tell
I guess not. In any case, like I said, your imports from dist will still work for types; esbuild discards them entirely before the part where the build script rejects "real" dist imports. We're also considering potentially moving those sorts of types to an external package, which would be an actual solution to the issue.

bin/index.mts
```diff
@@ -276,6 +276,16 @@ async function buildPlugin({ watch, noInstall, production, noReload }: Args): Pr
       build.onResolve({ filter: /^replugged(\/\w+)?$/ }, (args) => {
         if (args.kind !== "import-statement") return undefined;
 
+        if (args.path.includes("dist")) {
+          return {
+            errors: [
+              {
+                text: `Unsupported import from dist: ${args.path}`,
```
asportnoy, line 283: Can you explain the alternative here similar to what we had before?

scripts/create-import-wrappers.mts
```diff
@@ -1,15 +1,35 @@
 import { writeFileSync } from "node:fs";
+import { basename, join } from "node:path";
+
+type LocationsRegistry =
+  | string[]
+  | {
+      [x: string]: LocationsRegistry;
+    };
```
asportnoy, line 8: Use Record<> instead here

colin273: Can't. TypeScript complains about a circular reference if I use Record, but not with this version.

bin/index.mts
colin273: Done

bin/index.mts
```diff
@@ -273,23 +273,30 @@ async function buildPlugin({ watch, noInstall, production, noReload }: Args): Pr
...
     setup: (build) => {
-      build.onResolve({ filter: /^replugged.+$/ }, (args) => {
+      build.onResolve({ filter: /^replugged(\/\w+)?$/ }, (args) => {
         if (args.kind !== "import-statement") return undefined;
 
+        if (args.path.includes("dist")) {
+          return {
+            errors: [
+              {
+                text: `Unsupported import from dist: ${args.path} Import from either the top level of this module ("replugged") or a top-level subpath (e.g. "replugged/common") instead.`,
```
lexisother, line 283:
```suggestion
                text: `Unsupported import from dist: ${args.path}\nImport from either the top level of this module ("replugged") or a top-level subpath (e.g. "replugged/common") instead.`,
```

lexisother (changes requested):

lexisother (approved):

PR #559, opened 2023-09-24
Slash command api bug fix

Tharki-God:
fixes the bug of Clyde becoming replugged after using a command with an ephemeral return message. 

also added handling for sub-commands and sub-command groups... 

well actually the types for it aren't complete and I can't seem to figure out how to type it with Dynamic typing for options, if someone else can do that then it would be great help


src/types/discord.ts
```diff
@@ -1,4 +1,9 @@
+import type { Channel, Guild } from "discord-types/general";
+import { CommandInteraction } from "../renderer/apis/commands";
+import { GetCommandOptions, RepluggedCommandResult } from "./coremods/commands";
 export enum ApplicationCommandOptionType {
+  SubCommand = 1,
```
FedeIlLeone, line 5: As the [official Discord doc](https://discord.com/developers/docs/interactions/application-commands#subcommands-and-subcommand-groups) says. Consequently rename all SubCommand to Subcommand.
```suggestion
  Subcommand = 1,
```

src/renderer/apis/commands.ts
```diff
@@ -14,15 +14,87 @@ import { ApplicationCommandOptionType } from "../../types";
...
 const logger = Logger.api("Commands");
 
+let RepluggedUser: User;
+
 interface CommandsAndSection {
   section: RepluggedCommandSection;
   commands: Map<string, AnyRepluggedCommand>;
 }
 
+declare class UserClass implements User {
```
FedeIlLeone, line 28: This is not needed. Just define the type of the module that you are waiting for below as `typeof User`.

FedeIlLeone (changes requested):

asportnoy (approved):

FedeIlLeone (approved):

PR #567, opened 2023-10-08
Fix command type import

ShadiestGoat:
This is a really tiny pr that fixes auto completion for commands. 

Basically, I was running into an issue when writing with `Injector.utils.registerSlashCommand` - it was just not doing type checking. However, when importing the types using a relative import, it just works. 

Why? I have no clue, just typescript being typescript or whatever. All I can say is that this fixes it. The fact that `src/types` doesn't just work could have wider implications on other parts of the codebase, but honestly I have no clue and nothing has come up yet so maybe not


asportnoy:
@ShadiestGoat I just pushed some various type updates on the v4.7.0 branch that might've affected this, please check if this is still needed.

ShadiestGoat:
hey @asportnoy I just managed to test this & yes, this is still needed!

asportnoy (approved):

PR #570, opened 2023-10-12
Theme Presets

12944qwerty:
Adds theme presets.

I've tested with a modification of the [Catppuccin](https://github.com/qwerty-mods/Catppuccin) theme. 

Presets are specified in the manifest
```json
{
    ...
    "presets": [
        {
            "label": "flavorname",
            "path": "relative/path/to/flavor",
            "default": false
        }
    ]
}
```


asportnoy (dismissed):
Naming needs changing per internal discussion a few weeks ago

lexisother (approved):

PR #571, opened 2023-10-13
SliderItem margin fix and types

FedeIlLeone:
- Fixed SliderItem's margin when there are markers w/o a note.
Before -> After

image
image
- Updated types for `channels`, `constants` (Endpoints and Routes can be a function), `flux` (more precise and with generics), `fluxDispatcher` (more precise and with generics), `messages`, `Loader` (new enum).
- Fixed the word plugin being used for themes after the security patch.
- Removed a leftover console log.

### DIFF
--- a/src/renderer/modules/components/FormItem.tsx
+++ b/src/renderer/modules/components/FormItem.tsx
@@ -40,25 +40,22 @@ export type FormItemType = React.FC<FormItemProps>;
 
 export default ((props) => {
   const { note, notePosition = "before", noteStyle, noteClassName, divider, ...compProps } = props;
+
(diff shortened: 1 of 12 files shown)


src/renderer/modules/components/FormItem.tsx
```diff
@@ -40,25 +40,22 @@ export type FormItemType = React.FC<FormItemProps>;
 
 export default ((props) => {
   const { note, notePosition = "before", noteStyle, noteClassName, divider, ...compProps } = props;
+
+  const noteStyleDefault = notePosition === "before" ? { marginTop: 8 } : { marginTop: 8 };
```
asportnoy, line 44: Unnecessary ternary?

asportnoy (approved):

PR #573, opened 2023-10-14
Utility classes

Penguin-Spy:
Implemented features:
- re-adding the `aria-controls` attribute to all TabBar components (allows for settings icons without :nth-child)
- an attribute for the current nitro theme: `html[data-nitro-theme="MINT_APPLE"]`
- data attributes for messages (author id, author is bot/webhook, author is self, etc.)
- adds a `.replugged` class to the html element for themes to detect which client they're on


src/renderer/coremods/utilityClasses/index.ts
```diff
@@ -0,0 +1,138 @@
...
+    type: () => React.ReactElement;
+  }>(filters.bySource(".content.id)"));
+
+  // the Message component isn't exported, so it must be extracted like this
+  const uninjectMessagesComponent = inject.after(MessagesComponent, "type", (_, res) => {
+    uninjectMessagesComponent();
+
+    const MessageListComponent = res.props.children.type;
+    const uninjectMessageList = inject.after(MessageListComponent, "type", (_, res) => {
+      const messagesArray = res.props.children.props.children[1].props.children[1].props
```
asportnoy, line 91: This line errored when I tried to view a channel. Please add handling to this so it won't crash.

src/renderer/coremods/utilityClasses/index.ts
```diff
@@ -0,0 +1,138 @@
...
+  nitroThemeClass();
+  await messageDataAttributes();
+
+  // generic stuff
+  const observer = new MutationObserver(addHtmlClasses);
+  observer.observe(html, { attributeFilter: ["class"] });
+}
+
+export function stop(): void {
+  inject.uninjectAll();
```
asportnoy, line 137: Remove the observer and at least some of the classes

asportnoy (changes requested):

src/renderer/coremods/utilityClasses/index.ts
Penguin-Spy: this got removed because the webpack search was made a lot easier, but i made sure the new one has adequate error checking.

src/renderer/coremods/utilityClasses/index.ts
Penguin-Spy: removed the observer, the `replugged` class, and the `data-nitro-theme` attribute. the rest of the changes get removed when the (no longer) injected elements are re-rendered.

DracoClaw:
been using the message classes for a while now and it's good. only change i did was adding `&& !message.interaction` to the is-webhook class to stop it marking command replies as webhooks

fedeericodl:
Superseded by #719

PR #580, opened 2023-10-23
Add esbuild overwrites

ShadiestGoat:
This PR adds the ability for plugin & theme developers to overwrite the replugged esbuild config. 
This is especially useful for when plugin developers need special configurations for their plugin to build - eg. adding font loaders for their css files. 

The reason its a function rather than just extra config values is to allow developers to write their own plugins without removing the current required plugins.

The reason I chose to name the extra config file `esbuild.extra.mjs` is to avoid editors like vscode from using interpreting a `esbuild.(m)js` file as the thing to use for building (it causes a lang server crash)

This only affects plugin developers - this does not need a new replugged version release, just a new npm package.

### DIFF
--- a/bin/index.mts
+++ b/bin/index.mts
@@ -12,6 +12,7 @@ import {
   rmSync,
   writeFileSync,
 } from "fs";
+import {cwd} from "process";
 import esbuild from "esbuild";
 import path from "path";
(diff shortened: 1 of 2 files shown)


colin273 (approved):
LGTM

PR #581, opened 2023-10-24
Fix Updater

### LINKED ISSUE #524: Limit number of parallel update requests
The Replugged API has a rate limit of 20 requests per second. However, the updater currently requests all updates at once ([relevant code](https://github.com/replugged-org/replugged/blob/615fb152593a892d0e0b02a2e0fb0af39d738d17/src/renderer/managers/updater.ts#L221-L224)). The check should be throttled to 15 requests per second (give some wiggle room) and properly handle 429s.

In the future, we should add a route to the backend for a bulk update check to avoid needing a lot of requests like this.

TheHellTower:
Fixes #524 ?

From:
image<br>To:
DiscordCanary_UPosDipF0a

### DIFF
--- a/src/renderer/managers/updater.ts
+++ b/src/renderer/managers/updater.ts
@@ -134,13 +134,16 @@ export async function checkUpdate(id: string, verbose = true): Promise<void> {
   }
 
   const newVersion = res.manifest.version;
-  const versions = [ newVersion.split('.'), version.split('.') ];
+  const versions = [newVersion.split("."), version.split(".")];
 
-  if(newVersion !== version &&
-    (versions[0][0] > versions[1][0] || 
-     (versions[0][0] === versions[1][0] && versions[0][1] > versions[1][1]) || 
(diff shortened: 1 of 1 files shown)


asportnoy (changes requested):
This is intended behavior, not a bug. v4.7.1-v4.7.4 were only released on NPM as they do not have any changes for Replugged itself.

TheHellTower:
> This is intended behavior, not a bug. v4.7.1-v4.7.4 were only released on NPM as they do not have any changes for Replugged itself.

But this way anyway it would not give any update that is lower.. It's better to avoid them than just watch them and be like "`I will not click and see it daily`"...

Anyway, what about: https://github.com/replugged-org/replugged/pull/581/files#diff-a7f1a071f6cfadfb23f14451a3b445c50624212b74b26a7ba27b4e80f0dac31aL222-L226

asportnoy:
We want to be able to downgrade versions though.

If you have unrelated changes you want to make please PR them separately.

PR #582, opened 2023-10-24
change the tryPort to use hostname.local (for wsl to work)

thatgurkangurk:
when you try to use the --watch flag in wsl, it cannot connect since the networks are *technically* separated. if you use hostname.local however, the wsl "container" can connect to the main windows machine.

this should not break any other systems since i tested it in linux, linux in wsl and on windows without issue

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
 
 settings:
   autoInstallPeers: true
--- a/bin/index.mts
+++ b/bin/index.mts
@@ -24,6 +24,7 @@ import { logBuildPlugin } from "../src/util.mjs";
 import { sassPlugin } from "esbuild-sass-plugin";
 import { fileURLToPath } from "url";
 import { AddonType, getAddonFolder, isMonoRepo, selectAddon } from "./mono.mjs";
+import { hostname } from "os";
 
 interface BaseArgs {
   watch?: boolean;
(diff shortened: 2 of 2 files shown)


pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
```
FedeIlLeone, line 1: Revert. You are most likely using a pnpm version (v8.6.0 or v8.6.1) which incremented the package version causing many issues. Update your pnpm to the latest version, which no longer has this breaking change.
```suggestion
lockfileVersion: '6.0'
```

FedeIlLeone (changes requested):

pnpm-lock.yaml
thatgurkangurk: alr i reverted it, should be good now

12944qwerty (changes requested):
This unfortunately isn't supported on Mac devices. You may need to add a check to see if it's mac or not to choose to use `.local` or localhost.

yofukashino:
```ts
{
    find: "RPCServer:WSS",
    replacements: [
      {
        match: "127.0.0.1",
        replace: (prefix: string) => `0.0.0.0`,
      },
    ],
  }```
  
  Include this patch in replugged with a toggle for this. 
  
  Also if possible make it an argument or env variable.  
  
  `process.env.REPLUGGED_ADDON_HOST ?? "localhost"`

PR #584, opened 2023-10-27
fix webpack

Penguin-Spy:
types are broken when assigning to `webpackChunk.push` ([here](https://github.com/Penguin-Spy/replugged/blob/acffec100afe8114311f7e16deead508ad79d85f/src/renderer/modules/webpack/patch-load.ts#L80)). after a cursory glance i couldn't see how to make TS not mad. the code itself works just fine.

replugged loads, all of the common modules/components fail to be found, and some plugins do load if their webpack searches happened to not break.


Penguin-Spy:
fixed typing, TS is no longer mad :)
(there's still some stuff in other files that I didn't touch)

Penguin-Spy:
closing in favor of #585

PR #585, opened 2023-10-28
Fix unswc even more

Tharki-God:
fixes some coremods and fixes the plain text patch for some coremods not loading due to race conditions or something.
also fixed some errors with plain text patch eval logic and now it logs out the error if any + makes it easy to find patches file in sources tab


src/renderer/modules/webpack/patch-load.ts
```diff
@@ -97,41 +98,41 @@ function loadWebpackModules(chunksGlobal: WebpackChunkGlobal): void {
...
+      void patchChunk(loadedChunk);
     }
   }
 
   patchPush(chunksGlobal);
   signalReady();
 }
 
-// Intercept the webpack chunk global as soon as Discord creates it
+// Intercept he webpack chunk global as soon as Discord creates it
```
asportnoy, line 135:
```suggestion
// Intercept the webpack chunk global as soon as Discord creates it
```

src/renderer/coremods/badges/index.tsx
```diff
@@ -37,17 +37,15 @@ const cache = new Map<string, BadgeCache>();
 const REFRESH_INTERVAL = 1000 * 60 * 30;
 
 export async function start(): Promise<void> {
-  const mod = await waitForModule<Record<string, BadgeMod>>(filters.bySource("getBadges()"));
-  const fnPropName = Object.entries(mod).find(([_, v]) => typeof v === "function")?.[0];
-  if (!fnPropName) {
-    throw new Error("Could not find badges function");
-  }
+  const mod = await waitForProps<{ BadgeSizes: Record<string, string>; default: BadgeMod }>(
```
FedeIlLeone, line 40:
```suggestion
  const mod = await waitForProps<{ BadgeSizes: BadgeSizes; default: BadgeMod }>(
```

src/renderer/modules/components/Notice.tsx
```diff
@@ -21,11 +21,10 @@ export type NoticeType = React.FC<NoticeProps> & {
   Types: typeof Types; // for backwards compat
   HelpMessageTypes: typeof Types;
   default: React.FC<NoticeProps>;
-}
+};
 
 const NoticeComp = await waitForModule<NoticeType>(filters.bySource("WARNING=0]"));
 const Notice = NoticeComp.default as NoticeType;
-Notice.Types = Types;
```
FedeIlLeone, line 28: Add this back for backwards compatibility

FedeIlLeone (changes requested):

asportnoy (approved):

FedeIlLeone (approved):

PR #586, opened 2023-10-28
15 req/s ?

### LINKED ISSUE #524: Limit number of parallel update requests

TheHellTower:
Fixes #524

### DIFF
--- a/src/renderer/managers/updater.ts
+++ b/src/renderer/managers/updater.ts
@@ -219,10 +219,18 @@ export async function checkAllUpdates(autoCheck = false, verbose = false): Promi
 
   logger.log("Checking for updates");
 
-  await Promise.all([
-    checkUpdate(REPLUGGED_ID, verbose),
-    ...addons.map((addon) => checkUpdate(addon.manifest.id, verbose)),
-  ]);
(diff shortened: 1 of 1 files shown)


asportnoy (changes requested):
This will only run 1 request at a time and delay between each one. We want to run as many as possible at a time until we hit 15 within a second, *then* delay it until it won't be over 15/s.

TheHellTower:
Review please 😔

asportnoy (changes requested):
This will still only request one at a time rather than doing them in parallel

TheHellTower:
> This will still only request one at a time rather than doing them in parallel

Oh so you didn't mean in a row but really all of them at the "exact" same time

asportnoy:
Yes, it should run as many as possible in parallel until it hits the limit and then throttle starting the next request until 1 second has passed since the first request.

PR #590, opened 2023-12-01
Fix missing wpRequire.c

asportnoy:
For some reason, wpRequire.c is removed at some point after startup, but the chunks are still all on it, so I just save the reference to it while we still have it.


colin273 (approved):
Works on your machine 👍

PR #591, opened 2023-12-07
Parallel requests (15 requests/s)

FrimeX:
In theory this should be the fix for the rate limit. ref. #586


FrimeX:
I'm clueless why Prettier and ESLint failed thru the checks. They checked fine locally, no idea why it doesn't want to flag it successful here. Please enlighten me if someone can explain it.

PR #596, opened 2024-01-23
fix(bin): fix esbuild.extra.mjs import path on Windows

Socketlike:


FedeIlLeone (approved):
Tested on Windows and Ubuntu. LGTM

PR #604, opened 2024-03-12
fixed slash command api

YofukashiNo:
Fixed Slash commands not showing up
Embeded slash command's icon into replugged renderer itself


12944qwerty (approved):

src/renderer/coremods/commands/index.ts
```diff
@@ -9,35 +11,135 @@ import { loadCommands, unloadCommands } from "./commands";
...
+        fetchState: { fetching: boolean };
+        result: {
+          sectionIdsByBotId: Record<string, string>;
+          sections: Record<
+            string,
+            { commands: Record<string, AnyRepluggedCommand>; descriptor: RepluggedCommandSection }
+          >;
+          version: string;
+        };
+        serverVersion: symbol | string;
```
FedeIlLeone, line 26: Can you make a type just for this? This exact record is copied and pasted for almost every function below, resulting in duplicate code.

FedeIlLeone (changes requested):
Just this

src/renderer/coremods/commands/index.ts
```diff
@@ -9,35 +11,79 @@ import { loadCommands, unloadCommands } from "./commands";
 const logger = Logger.api("Commands");
 const injector = new Injector();
 
-interface ApplicationCommandSearchStoreMod {
-  useDiscoveryState: (...args: unknown[]) =>
+type commandState =
```
FedeIlLeone, line 14:
```suggestion
type CommandState =
```
A general convention in TypeScript is to use PascalCase for types/interfaces. Also replace the name below.

FedeIlLeone (changes requested):

FedeIlLeone (approved):

PR #606, opened 2024-03-28
Handle extant but invalid potential discord install dirs

### LINKED ISSUE #605: Plug fails if empty discord install locations exist on linux
### Describe the bug

If any directory in `KnownLinuxPaths` exists, it is assumed to be valid. In my case, an old backup of app.asar was not removed when a different discord package was uninstalled, meaning `/usr/share/discord` existed, containing only the app.asar_backup and no other files. This causes the installer to fail completely.

### Reproduction steps

1. Install the official discord client to somewhere like /opt/discord
2. Create a directory that's higher in KnownLinuxPaths, such as /usr/share/discord
3. Attempt to `pnpm run plug`
4. Installer will fail with the message "Failed to find Discord path"

### Expected behavior

`findPathFromPaths` should handle the case of an extant, but invalid, potential discord install directory, and continue to check the other potential locations, possibly with a warning that an invalid discord installation was found.

### Actual behavior

When `findPathFromPaths` is called, it iterates over `KnownLinuxPaths` in order, and since `/usr/share/discord` (the old invalid directory) exists and is above `/opt/discord` (the correct discord directory), `findAppAsarInDir` is called on the invalid directory and returns null, causing the installer to fail completely.

### Additional information

_No response_

KatTheFox:
fixes #605 

The warning message is there for cases like mine, where the old install remnants are unwanted. However, it may cause concern to other users who don't care or know what it means. I personally believe that's worth the risk, as this patch only affects linux users who tend to be less scared of the command line overall, but the message will be removed if the maintainers disagree.

### DIFF
--- a/scripts/inject/platforms/linux.mts
+++ b/scripts/inject/platforms/linux.mts
@@ -32,8 +32,8 @@ const findPathFromPaths = async (
   platform: DiscordPlatform,
   paths: string[],
 ): Promise<string | null> => {
-  let discordPath = paths.find((path) => existsSync(path));
-  if (!discordPath) {
+  let discordPaths = paths.filter((path) => existsSync(path));
+  if (discordPaths.length === 0) {
     const readlineInterface = readline.createInterface({
       input: process.stdin,
       output: process.stdout,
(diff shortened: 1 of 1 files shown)


FedeIlLeone (approved):
Works as expected, tested on Ubuntu 22.04 with multiple installations of the app.

PR #613, opened 2024-04-09
Notification API

YofukashiNo:
An API for in app notification


yofukashino:
An example

```js
const notif = replugged.notification.NotificationAPI.coremod("Notification")
notif.notify({
    timeout: 60000,
    type: "danger",
    header: "DANGER",
    gradient: ["#FF0000", "#F00000"],
    content: "CLOSE YOUR CLIENT YOU ARE GETTING TRACKED!"
})
```

PR #614, opened 2024-04-09
slash command ephemeral message fix

yofukashino:


src/renderer/apis/commands.ts
```diff
@@ -26,10 +26,10 @@ interface CommandsAndSection {
   commands: Map<string, AnyRepluggedCommand>;
 }
 
-void waitForModule<typeof User>(filters.bySource(".isStaffPersonal=")).then((User) => {
+void waitForModule<typeof User>(filters.bySource('"isStaffPersonal",')).then((User) => {
   RepluggedUser = new User({
     avatar: "replugged",
-    id: "replugged",
+    id: "69",
```
lexisother, line 32: This change is no longer required. I got it working just by replacing the source match.

src/renderer/coremods/commands/index.ts
```diff
@@ -270,7 +270,8 @@ async function injectProfileFetch(): Promise<void> {
     fetchProfile: (id: string) => Promise<void>;
   }>("fetchProfile");
   injector.instead(mod, "fetchProfile", (args, res) => {
-    if (args[0] === "replugged") {
+    //69 is replugged user id
+    if (args[0] === "69") {
```
lexisother, line 274: No longer required.

lexisother (dismissed):

src/renderer/apis/commands.ts
YofukashiNo: should i close the pr then?

lexisother: No? The source match change is still required.

lexisother: Yes, I also saw that, but you have to consider that was quite a while ago. I tested it *today* without replacing the `id` and it worked fine. However, I didn't test clicking the fake user. Is that where the issue lies?

YofukashiNo: yes

lisekilis (changes requested):
Change from `69` to `826980768571716968` aka REPLUGGED in ascii (with the separators removed)
... also fix prettier lol

colin273 (approved):

PR #630, opened 2024-06-19
Dependencies updates

FedeIlLeone:
- Updated most dependencies to their latest version, and fix all vulnerabilities (#629 , #628).
- Updated all actions to their latest version available.
- Updated node and chrome versions in build scripts to match Discord's.
- Downgraded node to v20. v21 is now in EOL, v20 is the version in LTS.


src/renderer/modules/common/hljs.ts
```diff
@@ -1,4 +1,4 @@
-import type HighlightJS from "highlightjs";
+import type HighlightJS from "highlight.js";
```
colin273, line 1: Why? Do we really need to bring in the full hljs library as a dependency, even though we're just using the types? That's the whole point of the types package.

colin273 (commented):
Mostly good, just not sure about the switch from the DefinitelyTyped highlightjs package to the full highlight.js module.

src/renderer/modules/common/hljs.ts
FedeIlLeone: I did it because I noticed that it is a package that has been deprecated for several years, remaining at v9 of hljs, because from v10 they brought all the types to the main package. Discord uses v11 of hljs and therefore the types were incorrect with the v9 types.

colin273 (approved):

PR #634, opened 2024-06-23
fix(scripts): catch and exit if building fails

FedeIlLeone:
Previously it would have printed an error in the console but would have continued anyway saying that the building process went ok. With this PR instead if an error is caught by the building then we exit immeditately with a failure code 1. In this way the jobs of a workflow on GitHub also fail if something goes wrong.

https://discord.com/channels/1000926524452647132/1000955967627874424/1254504021570486302
The job would have ended successfully even if the building had not gone well, making the developer believe that the addon was ready.

Tested with various cases, even with watch.

### DIFF
--- a/bin/index.mts
+++ b/bin/index.mts
@@ -272,7 +272,7 @@ async function handleContexts(
       if (watch) {
         await context.watch();
       } else {
-        await context.rebuild().catch(() => {});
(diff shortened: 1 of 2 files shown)


colin273 (approved):
LGTM

PR #636, opened 2024-07-06
Fix crashes

colin273:
Discord changed their webpack configuration a little bit, so that most of the common modules we need are now declared inline rather than pushed to the global array of chunks. This update fixes our injection to account for this change. It also includes fixes for the plaintext patches in our coremods, so that if initialization fails, missing coremods in the `replugged.coremods.coremods` namespace will not cause a React crash.

In the process, I've also swapped out the wait/signal start mechanism for a simpler means of intercepting the creation of Webpack chunks/modules. It adds needless complexity and is likely complicit in our perennial crashing issues. So far, with this mechanism removed, Replugged loads reliably on my machine. More testing is needed, especially on Windows and Linux, to confirm these stability improvements.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -62,7 +62,8 @@
     "Vendicated",
     "webauthn",
     "weblate",
-    "Withs"
+    "Withs",
+    "YofukashiNo"
   ],
   "ignoreWords": [],
   "import": [],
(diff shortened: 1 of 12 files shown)


FedeIlLeone (approved):
LGTM

yofukashino:
Leaving this comment as a reminder for the issue: Plaintext patch for all plugins not being loaded

PR #644, opened 2024-08-31
custom titlebar on linux

yofukashino:


src/renderer/coremods/settings/pages/General.tsx
```diff
@@ -110,6 +114,21 @@ export const General = (): React.ReactElement => {
...
       </SwitchItem>
 
+      {
+        // TODO: i18n
+        DiscordNative.process.platform.includes("linux") && (
+          <SwitchItem
+            value={titlebarValue}
+            onChange={(value) => {
+              titlebarOnChange(value);
+              restartModal(false);
```
FedeIlLeone, line 124:
```suggestion
              restartModal(true);
```
A full app restart is required to load the custom titlebar.

FedeIlLeone (changes requested):
Works as expected, tested on Ubuntu 22.04. I'll add the strings for i18n and everything else is ok except for that restart modal.

FedeIlLeone:
> Works as expected, tested on Ubuntu 22.04. I'll add the strings for i18n and everything else is ok except for that restart modal.

Added strings and changed the restart modal to do a full relaunch.
I also fixed a bug with overlays that added the title bar there. removing an unnecessary plaintext patch fixes this.

PR is ready for v4.9.0.

FedeIlLeone (approved):

PR #645, opened 2024-09-07
Async Injections

yofukashino:
Adds an option to patch inject in functions with async callbacks.

This working POC and so obviously, a lot of refactoring might be possible. Hence if you may please help me with it or guide me with it.


yofukashino:
closing this as I think iterating  on the instead function would be better than async for all three

PR #651, opened 2024-10-20
refactor: context menu api

FedeIlLeone:
This PR rewrites a great part of the Context Menu API (coremod), allowing for a more readable code and less type casting. This fixes the now year old bug with using the useState React hook with menu items (for example MenuCheckboxItem), that has blocked a few plugins from updating on the store. Types on its components have been updated to the latest app build.

~Keeping it a draft PR until a few more tests on various context menus of the app.~


colin273 (approved):
Resolves the server context menu crash for me (tested with SHC). Should be good to go.

PR #653, opened 2024-11-11
Linux Improvements

### LINKED ISSUE #518: [Linux - Arch] Leftover files if discord is uninstalled before unplugging
### Describe the bug

# Not looking for a fix at the moment, this issue is more to just have this issue on record.
if discord is removed via the package manager before unplugging, 2 files will be left behind.
`/opt/discord-[development/canary/ptb]/`
|- `app.asar`
|- `app.orig.asar`

A quick patch would be to add some sort of `pnpm run cleanup` to remove leftover files, as this can prevent discord from reinstalling cleanly, ex:
```
errors occurred, no packages were upgraded.
error: failed to commit transaction (conflicting files)
discord-development: /opt/discord-development/resources/app.asar exists in filesystem
Errors occurred, no packages were upgraded.
 -> error installing: [discord-*version*.tar.zst] - exit status 1
 ```

### Reproduction steps

1. Remove discord (do not unplug prior)
2. Remnants at `/opt/discord-[edition]

### Expected behavior

There should be no remnants, or at the minimum, no conflicting files

### Actual behavior

conflicting files are left behind

### Additional information

_No response_

yofukashino:
- fixes https://github.com/replugged-org/replugged/issues/518
- no longer stops package managers from updating discord (gets unplugged instead)
~~fixes flatpak permission issue, including this since https://github.com/replugged-org/replugged/pull/625 is stale for a while~~
- added replugged config even if using dev install for flatpak


yofukashino:
this would break previous installs and would require a manual unplug with older version and plugging with this

yofukashino:
moving the original asar in app.asar breaks windows squirrel updater. any ideas for fix?

yofukashino:
Moved elevate from import to function.
Now elevates only if writing to root installs not `/home` 
Also only after deps are installed

PR #654, opened 2024-11-11
Fix i18n (mod crashing)

FedeIlLeone:
Fixes the mod crashing on startup.

- Huge refactor on Replugged's i18n to adapt to the new `discord-intl` system: uses two build tools present on the discord-intl repo , that have been rewritten for esbuild, to load and transform the i18n messages to the new format readable with the formatters. This also comes with a types generator, which let's us type our i18n strings perfectly. i18n files are no longer in the bundled asar but instead compiled with the renderer.
- Added support for `es-419` (Español, LATAM).
- Fixed plaintext patches.
- Fixed an issue with toasts not being displayed in case an addon fails to enable or disable, due to missing strings or non existent variables.
- Fixed an issue with the watching mode of an addon which would've showed both success and failed toasts in case the reload failed.
- Removed `node-fetch` and `@octokit/rest`, reducing the size of the built main.js by ~60% (~650kb to ~250kb).
- Updated dependencies.


FedeIlLeone:
🚧 TODO: remake the API to add strings to Replugged's i18n for plugins.

lexisother (approved):

Issue #612, opened 2024-04-08
Voice API/Preload API


YofukashiNo:
### Describe the feature request

An API for devs to record audio (with user permission obv, probably a replugged setting)
An alternative idea would be to give plugins preloader access to some extent as in something like `replugged.plugin.getPluginPreload(pluginId)` / `RepluggedNative.plugins.getPluginPreload(pluginId)` and add a different warning during installation for plugins if manifest mentions preload. I mean client mod should at least give them the option to have it, should be up to users for their own safety if they install malware option

### Why should this be added?

voice message plugin or literally any plugin that might record audio.
but Preload would be useful for other things as well such as predefined user directories for saving images or even more

### Please confirm the following

- [X] This is a feature request for Replugged itself, not a feature request for a plugin or theme
- [X] [I searched the issues](https://github.com/replugged-org/replugged/issues) to make sure this feature has not been requested before.

PR #656, opened 2024-12-06
Native Api

yofukashino:
- Adds https://github.com/replugged-org/replugged/issues/612

Just a concept draft if you may even wanna entertain the idea of adding this.


yofukashino:
This pr can be tested/tried out with https://github.com/yofukashino/YoutubeAdblock

Issue #617, opened 2024-05-07
Wrong Source from react devtools


yofukashino:
### Describe the bug

When a component type is injected from the injector and another developer attempts to view the component's source from React DevTools, it redirects to Replugged's injector instead of the component itself. This can pose a challenge for developers trying to access the component's source.

### Reproduction steps

1. Inject into component
2. Find Component in React Tree
3. Click View Source

### Expected behavior

Redirect to component's source

### Actual behavior

Redirect to injector's source

### Please confirm the following

- [X] I checked if Discord and Replugged are up to date
- [X] I made sure this is not related to an external plugin or a theme
- [X] [I searched the issues](https://github.com/replugged-org/replugged/issues) to make sure this bug has not been reported before.

PR #660, opened 2025-01-22
rdtComponentSourceFix

yofukashino:
Fix for https://github.com/replugged-org/replugged/issues/617


src/renderer/coremods/rdtComponentSourceFix/index.ts
```diff
@@ -0,0 +1,25 @@
+import { AnyFunction } from "src/types";
+
+function getOriginal(fn?: AnyFunction): AnyFunction | undefined {
+  return Function.toString.apply(fn).includes("objInjections.injections.get(funcName)")
```
FedeIlLeone, line 4: This function expects `fn` to be always of type 'function'. Sometimes this is not the case, for example when starting up the coremod the first time, `fn` will be undefined. Adding a check and returning `fn` if the typeof of `fn` is not a function is enough.

FedeIlLeone (changes requested):

FedeIlLeone (approved):

PR #663, opened 2025-01-27
feat: disable sentry

yofukashino:
Disable loading sentry and posting to metrics/science completely.


FedeIlLeone (approved):
LGTM, I've added the block of the sentry.*.js file for other clients as well.

PR #664, opened 2025-01-29
Components fix

yofukashino:
^^_


FedeIlLeone (approved):
LGTM, I changed a few source filters because they were getting the wrong component.

PR #671, opened 2025-02-18
I18n Fix

yofukashino:
Updates `@discord/intl` and some types related to it

Still crashes on reload for me but that is probably ignition issues


fedeericodl:
Closing in favor of #678

PR #672, opened 2025-03-03
chore: update to eslint 9

fedeericodl:
This pull request includes a major version upgrade to ESLint and some minor code improvements. The primary focus is on migrating to the new ESLint flat config, updating dependencies, and improving TypeScript type safety. Thanks to this change we leave the config dependency we have relied on in these years, now allowing greater control on it.

* Removed old ESLint configuration files (`.eslintignore` and `.eslintrc.yml`) and replaced them with the new standard flat config with `eslint.config.mjs`. Most of the rules from the dependency have been ported to the new config to keep the current style of the repository as close as possible before this change. This new configuration includes recommended settings for ESLint, TypeScript, and React (added new), along with specific rules for various file types. In some specific cases comments have been added to ignore some linter rules, most of them are about `no-unnecessary-condition`. It's better to keep it enabled and disable it occasionally in very specific cases where it is necessary to keep the condition unnecessary.
* Updated dependencies, and order the keys properly in the `package.json`. The package manager version was also updated from pnpm 9.x to pnpm 10.x.
* Improved type safety in various scripts, such as `scripts/build-bin.mts`, `scripts/inject/checks/env.mts`, and `scripts/inject/index.mts`.
* Minor refactoring and cleanup of code to improve readability and maintainability.


fedeericodl:
Need to wait for a new node.js v20 version that fixes corepack issues. It should be out in a few days.

fedeericodl:
Updated node and checks passed. PR out of draft.

colin273 (approved):

12944qwerty (approved):

PR #680, opened 2025-04-08
feat: add support for desktop visual refresh

fedeericodl:
- [5260c639386e4f6d0a5cd05b4af2f451bcf35b30] Styles now use the new spacing variables to be more consistent with everything else. This also adds support for UI density, specifically how much space there is between each addon in the list. Addon cards now have a subtle border, which is now applied in all cards in the app, again for consistency and for greater distinction when in light or onyx theme.
image
image
- [22ceefff8b7d4fe7cf26dfc326f1b3344f9a7595] Patch the "hasNotice" function to handle style changes if notices are present; class names are added to the outer frame to add rounded edges or not.
- [a6575cad2e33097aff0969bb7c7245ce229d5632] Update duration formatting in updater settings to use localized strings.
image

### DIFF
--- a/src/renderer/coremods/notices/index.tsx
+++ b/src/renderer/coremods/notices/index.tsx
@@ -12,7 +12,7 @@ function Announcement({
   onClose,
 }: RepluggedAnnouncement): React.ReactElement {
   return (
-    <Notice className="replugged-notice" color={color}>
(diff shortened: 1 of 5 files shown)


lexisother (approved):
LGTM

PR #701, opened 2025-06-30
Add Universal Theming Classes

yofukashino:
Adds UTC (Universal Theming Classes) that appends stable class names [For example `utc_xyz`] to elements, so theme devs don’t have to worry about class rerolls, only structural changes.
Helps with theming by removing the need for wildcard selectors, which also improves performance.

Named it UTC (on the recommendation of @JustOptimize) with the idea of getting it adopted across other client mods too, to make theming easier and consistent.


JustOptimize:
NGL using UTC as the name was a joke cause Coordinated Universal Time should have been CUT anyways and asking people to enable/implement UTC would have made some funny confusion.

JustOptimize (approved):

PR #704, opened 2025-07-06
feat: settings.useValue

yofukashino:
Use a setting's value in React, and let it update even if updated from outside of the component. Useful for updating the component in the main UI as the setting changes from the settings's menu


src/renderer/apis/settings.ts
```diff
@@ -66,6 +68,34 @@ export class SettingsManager<T extends Record<string, Jsonifiable>, D extends ke
     return this.#settings[key] ?? fallback ?? this.#defaultSettings[key];
   }
 
+  /**
+   * Use a setting's value in react.
+   * @param key Key of the setting to retrieve.
+   * @param fallback Value to return if the key does not already exist.
+   * @returns
+   */
+  public use<K extends Extract<keyof T, string>, F extends T[K] | undefined>(
```
lexisother, line 77: Would prefer `useValue`.

fedeericodl (approved):

PR #709, opened 2025-07-15
feat: vanilla

yofukashino:
pass `--vanilla` to not load replugged at all. 

I did replugged stuff into seperate instead of `index.ts` because a I didnt want to wrap everything by a single if else


yofukashino:
To be Blocked by https://github.com/replugged-org/replugged/pull/741

PR #723, opened 2025-07-28
feat: addon header buttons redesign 

yofukashino:
The image explains it all
image

### DIFF
--- a/src/renderer/coremods/settings/pages/Addons.css
+++ b/src/renderer/coremods/settings/pages/Addons.css
@@ -47,11 +47,6 @@
   color: var(--header-primary);
 }
 
-.rp-addon-header-buttons {
-  align-self: normal;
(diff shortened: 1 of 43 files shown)


src/renderer/coremods/settings/pages/Addons.css
```diff
@@ -46,3 +46,12 @@
 .replugged-addon-breadcrumbsInactive:hover {
   color: var(--header-primary);
 }
+
+.rp-addon-header-buttons {
+  align-self: normal;
```
fedeericodl, line 51: You don't need this if you remove `align={Flex.Align.START}` from the parent `Flex`. The align START is no longer necessary, can be removed, style does not change with or without an open addon settings.

fedeericodl (changes requested):

PR #733, opened 2025-08-15
feat: React Error Decoder Coremod

yofukashino:
New coremod for decoding React errors. Instead of a link to the full error, the error itself.
Including the JSON is much easier than getting the JSON on every launch. 

Currently, Discord uses React `19.0.0`. The error code JSON can be downloaded from
`https://raw.githubusercontent.com/facebook/react/v${versionNumber}/scripts/error-codes/codes.json`


lexisother (changes requested):
I will not be approving this PR if the error code file is bundled. You have every resource available to you to fetch this dynamically and cache it if needed, even getting the correct file version based on the React version in use by Discord.

Until this is improved, it's a no from me. No discussion.

lexisother:
Proxying from the website seems overkill. Besides, I am fairly confident in the fact that fetching it on launch will not result in a 429 unless you are restarting your client more than 5000~ times per hour.

yofukashino:
> Proxying from the website seems overkill. Besides, I am fairly confident in the fact that fetching it on launch will not result in a 429 unless you are restarting your client more than 5000~ times per hour.

I have gotten 429 on it before multiple times :<

PR #757, opened 2025-10-15
Splash presets

yofukashino:
- Added key for `splash` in theme presets
- Changed the `path` key to `main` for main client preset
- Made both of the keys optional in manifest
- Added logic to load theme based on preset in splash
- Fixed the issue of presets not using `esbuild.extra` overwrites
- Added key `id` for preset id and made it optional (label can be used)
- switched `outdir` to `outfile` for presets, example `replugged://theme/.../presets/.../main.css`
- Setting for preset now uses `id` instead of `path`


fedeericodl (changes requested):
We've already talked about this in chat, don't split the selection of theme presets for the main and splash. The current select must control both splash and main.

fedeericodl (approved):

